PR #235, opened 2022-04-29
Add Docker Components for Staging Env

### LINKED ISSUE #78: Research Viability and Implementation of Staging Website
### Overview
As a developer, we need to ensure that our work is easy for the team to review and comment on. For this issue we will conduct all the work needed to create a staging version of our product.

### Action Items

- [x] Research uploading the website onto a host/server
  - [x] Also research ways to have the repo (this repo) selectively upload to the staging or production website
- [x] Create a configuration proposal in the form of code
  - [x] This includes configurations as part of our frontend, backend, and database
  - [x] This also includes configurations based on infrastructure as code
  - [x] The proposal must make security as its top priority
    - [x] As part of this research how passwords to our various servers are encrypted, and how to prevent client access; make sure that the tools used for our servers provides this kind of protection
- [x] Find out whether we can create a "https://staging.civictechjobs.org" website with the civictechjobs.org domain alone
- [x] Implement the staging website
  - [x] Be sure to discuss with the ops team on how our env variables are protected/encrypted to prevent leaking it out
- [x] Edit the pull-request-template.md to include photos of visual changes.
- [ ] Update documentation on devOps details
- [x] Release
  - [x] #189 

### Resources/Instructions
[Link to current deployment plan](https://github.com/hackforla/CivicTechJobs/issues/78#issuecomment-1033233215)
[OWASP Top Ten](https://owasp.org/www-project-top-ten/)
[White-noise](http://whitenoise.evans.io/en/stable/)

The Ops team at HackForLA manages hosting our various projects. Please make contact with them through Slack under the #ops channel. The stack that they are using are Terraform and AWS, so focus research on those tools.

Concerning the selective upload, GitHub natively have ways of doing this. The slightly more dangerous way is to have two GitHub actions, which runs on dispatch to either the staging or production environment. This is potentially dangerous since reverting is difficult if the the wrong dispatch is accidentally ran without a history of which commit was previously hosted. The better way would be to link it to [Github release ](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository)system, since releases are very purposeful. The final solution might involve a bit of both, but please do broad research on possible options!

Enzyme3:
Fixes #78 

### Changes
- Enable [whitenoise](http://whitenoise.evans.io/en/stable/) middleware
- Add gunicorn as requirement
- For stage environment:
  - Add stage directory
    - Removed stage directory from .gitignore
      - I don't believe filtering these is required this is required, as all it contains is a Dockerfile + example .env file
  - Add Dockerfile, .env, and docker-compose file for stage env
  - use Djangno with gunicorn in stage


Enzyme3:
@Aveline-art , this is the initial stab at the Docker setup for staging. Can you take a look? I'd like to send it over to the ops team for review.

Additionally, I'm currently stuck with a remaining [linter rule](https://github.com/hadolint/hadolint/wiki/DL3013) in `./stage/Dockerfile`. I used the Dockerfile you gave me as a template, where a multi-stage build is used such that a builder builds wheels from the `requirements.txt`, which is then copied by the runtime image for the actual installation. But I'm struggling to get this setup to work alongside the mentioned linter rule.

LMK what you think.
We can either:
1) [disable the rule](https://www.containiq.com/post/hadolint) for that line
2) removing the multi-stage build for the python lib installation
3) fix it somehow without resorting to either of those two approaches (TBD?)

Aveline-art (changes requested):
Everything looks very good! A couple of points:

1. Let's separate out the staging specific python dependencies into a separate file, maybe requirements_stage.txt. There might be some overlap but that is okay.
2. `stage/super-linter.log` seems to be an unnecessary files that got into the commit. If this is not needed, let's remove it.
3. When I built and ran the files, an `app/staticfiles` directory was created. Can this be added to .gitignore so contributors would not accidentally commit them?
4. For the linter issue, I am okay with option 1 or 2. The path of least-resistance is probably to disable the rule via inline comment.

Aveline-art (approved):
Everything looks good!

PR #259, opened 2022-05-20
Add pre commit for python linter and formatter

### LINKED ISSUE #258: Setup linter / formatter for python
### Overview

We need to ensure that all backend developers who are contributing to the project are committing code that is stylistically consistent and follows general best practices. The usage of linters and formatters will help enforce this consistency.

### Action Items

- [ ] Determine linters / formatters to be used
- [ ] Determine how the linters / formatters will be triggered
- [ ] Implement

### Resources/Instructions

- [Super-Linter](https://github.com/github/super-linter) is already setup as a Github action, but its use has been disabled for python files. Based on my current understanding, I am not certain if we should solely rely on this because:
  - It can suggest fixes, but [cannot automatically apply them](https://github.com/github/super-linter#how-it-works), which makes it suitable for linting, but not really for formatting
  - [Cannot natively install python dependencies](https://github.com/github/super-linter/issues/157), which results in lots of import errors for django modules. Technically, there appears to be a [workaround](https://github.com/github/super-linter/issues/157#issuecomment-648850330) but sounds kludgey and brittle 
  - bad developer UX. Instead of listing all of the lint errors, the errors are displayed on a per-file basis, i.e.  with lots of boilerplate text in between each file. Have to do lots of scrolling to view the relevant messages
  - slow. even after container is up, much slower than one would expect for a linter

Enzyme3:
Fixes #258 

### Changes
- Added a [pre-commit](https://pre-commit.com/) hook
- Added `app/.pre-commit-config.yaml` which:
  - formats with [isort](https://pycqa.github.io/isort/) on python files
  - formats with [black](https://github.com/psf/black) on python files
  - lints with [flake8](https://flake8.pycqa.org/en/latest/) on python files
  - lints with [pytlint](https://pylint.pycqa.org/en/latest/) (with pylint-django plugin) on python files
    - ignores the `mkdocs` directory
  - formats miscellaneous things like whitespaces and EOF for all files
- Applied the formatting and addressed the linting errors on the existing codebase

Lots of files were modified due to the formatting/linting but those were purely cosmetic. Should focus more on:
- `app/.pre-commit-config.yaml`
- `app/setup.cfg`


Enzyme3:
Lint error is: `/github/workspace/dev/webpack.dockerfile:1 DL3007 warning: Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag`

the formatter removed some whitespace and it looks like changing the file pulled it in for validation by super-linter. It was originally checked in as `node:latest` and that has not changed. LMK if you want to just suppress it or pin it to a particular version

Enzyme3:
added README.md under app on how to setup linter/formatter

app/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Setup Local Environment
+- Clone this repo
+- Navigate into this directory
+  - `cd ./CivicTechJobs/app`
+- Setup your python virtual environment and install dependencies
+  - `python -m venv .venv`
+  - `source .venv/bin/activate`
+  - `pip install -r ./requirements.txt`
+    - To deactivate virtual environment, type `deactivate`
```
Aveline-art, line 26: I am afraid that having a python venv would make having docker for dev a tad redundant (that said having `node_modules` outside of the container is also redundant). I would actually like for all dependencies to eventually move into the container, if possible.

Can the linter run via docker-compose in some way, such as through a dedicated linting container and the commit hook would run the docker-compose command before code commit?

Aveline-art (commented):
Hi @Enzyme3 . Great work on this as always.

A few more notes:

- The `readme` should be placed inside of our installation instructions which would help ensure our documentation is all in the same place rather than in multiple places at once.
- For the lint error, node:18 is fine as far as I can tell.
- If these changes are not critical for setting up our staging env, let's merge this into our `dev-branch-1` branch to avoid creating too many merge conflicts. Also it looks like your working branch has not been updated in a while, just fyi :)

app/README.md
Enzyme3: Thanks @Aveline-art  !
* deleted new `README` and moved contents to mkdocs section
* changed to node:18
* set new merge branch destination
* moved all linting dependencies to container that can now be invoked via `docker-compose` command. Have provided updated steps in `CivicTechJobs/mkdocs/docs/developer/installation.md`
  * this new approach uses the pre-commit.com framework but it does not create the commit hook quite yet. I think we can set this aside as a later TODO
  * also, hadolint (which is run in GH actions) is not yet incorporated into this container either. Have provided workaround in the doc, but I think we can set this also aside as a TODO as well

Aveline-art (approved):
Hi @Enzyme3 . Thank you for waiting for so long. I had to wait for the issue to be approved before merging. Everything looks perfect! Just a few small changes:

- Let's move the .env file into the `dev/` directory to keep the env with the related dockerfiles. Likewise, we can also rename the file to linter.env or some such. Make sure that linter.dockerfile still works
- Since we have linted those files, let's take out these lines from linter.yml

```yml
  VALIDATE_PYTHON: false
  VALIDATE_PYTHON_BLACK: false
  VALIDATE_PYTHON_FLAKE8: false
  VALIDATE_PYTHON_ISORT: false
  VALIDATE_PYTHON_MYPY: false
  VALIDATE_PYTHON_PYLINT: false
```

Once that is done, feel free to merge it at your leisure.

Enzyme3:
Thanks @Aveline-art , have renamed the .env file and placed it in `dev`
I did not take out the python references in the `linter.yml` yet because we probably still want to have the linter auto-execute everytime the PR gets created. 

Perhaps another TODO can be raised to enable it, and ensure that the local linting rules match with the github one

PR #318, opened 2022-10-12
svg descriptions added

gjagnoor:
Issue link - https://github.com/hackforla/CivicTechJobs/projects/1#card-85037168

Fixes - 1 file changed technically speaking, all other changes are from updating dev-branch-1 on my local machine with the updated version (changes others have made to it)

### Changes

added descriptions for the SVG images in app/frontend/src/assets/images/images.tsx using desc prop

descriptions - 
- cop-datascience-icon  - icon representing the close button
- cop-icon-engineering  - an engineering icon that represents the terminal  
- cop-icon-ops - an operations icon that represents a running operation being managed  
- cop-icon-produce - a product icon that represents a team communicating with each other  
- cop-icon-uiux - an icon for uiux representing a variety of ideas  
- icon-down-arrow  -     an icon representing the arrow pointing down
- icon-hamburger-menu -     the three parallel lines - hamburger menu icon 
- icon-x  -     icon representing the close button
- landing-page-bg -     landing page graphic design background - beige color
- landing-page-fg -     landing page main graphic design 
- horizontal-logo-on-dark -     a simple black rectangle - horizontal
- logo-horizontal -     civic tech jobs horizontal logo
- logo-logomark -     civic tech jobs logomark
- logo-logotype -     civic tech jobs - the name of the platform typed out
- logo-stacked-on-dark -     a small black rectangle
- logo-stacked - civic tech jobs logo and stacked platform name
- logo-vertical-     logo and platform name - vertically aligned
- logo-wordmark -     platform name and logo with a wordplay

PR #319, opened 2022-10-12
Webpack improvements

gjagnoor:
https://github.com/hackforla/CivicTechJobs/projects/1#card-74033514

`Tree-shaking` implementation is complete and ready for review. 

`Minification` seems to be already implemented. `webpack4+` minifies the bundle by default so we don’t need to implement it for `production`. `Should I mark this improvement as implemented in the ticket then?`

`Lazy loading` implementation is complete [along with `code-splitting` which was a prerequisite to this feature]

Fixes - 3

### Changes

- `tsconfig` setups the project to use `esnext modules` instead of `commonJS`. commonJS export import statements are not compatible with web pack tree-shaking. Is commonJS central to our app, or is this change in `tsconfig` ok?  
- added tree-shaking support to `webpack config` under `optimization`. See `usedExports`
- image showing tree-shaking support - 
Screen Shot 2022-10-12 at 3 49 42 PM

 - lazy loaded the landing page intro as an example for now. When you refresh the app you'll see `...loading` appear on the top left side which is an indication the component was lazy loaded. 

Please Note - In order for lazy loading to work, components have to be exported by default. I made the changes for LandingPageIntro and LandingPageCOP


Aveline-art:
Going to close this PR for the same reason as stated [here](https://github.com/hackforla/CivicTechJobs/pull/318#pullrequestreview-1139801546).

PR #324, opened 2022-10-13
Selection feature

### LINKED ISSUE #307: Create hook for drag selection of calendar
### Overview

As a developer, we need to fix bugs with our code. For this issue, we will create a hook that fixes the selection for the calendar page.

### Action Items

- [ ] Create a custom hook that detects when a user has their mousedown, and hovering over the cells of the Calendar component.
  - [ ] If the mouse is down, and the mouse is hovering over a cell, that cell should swap colors, from blank to green or green to blank.

### Resources/Instructions

gjagnoor:
resolves #307 

### Changes

`Calendar.tsx` - I moved a lot of code from `Calendar.tsx` file to hooks file in `Utils/hooks/dragToSelectUnselect.ts`. 

`dragToSelectUnselect.ts` - The hook and functionality are fully functional. A good next step in the future would be to condense this hook even more so it's easier to use for others in the future. But since it's not a top priority, I'll move onto other things for now. 

Here is what it looks like : 

Select/Deselect feature: 

dragtoselectunselect


app/frontend/src/components/Inputs/Calendar.tsx
```diff
@@ -145,55 +141,40 @@ function CalendarRow(props: CalendarRowProps) {
...
   useEffect(() => {
-    props.onChange(isSelected);
-  }, [isSelected]);
-
-  function handleClick() {
-    setIsSelected(!isSelected);
+    setData(nextdata)
+  }, [next])
+
+  function mouseDown(e: any) {
```
Aveline-art, line 151: I think we would want to keep it to work on click and on drag.

Ideally, it should work just like the [when2meet calendar](https://www.when2meet.com/?17722526-XW6jr).

Aveline-art (changes requested):
Hi @gjagnoor! Thank you so much for working on this. Just a small note. Sorry it took me a while to get to this.

Also, I did found a bug where if my mouse is released, the drag to color still works as if the mouse is still down. I'm not sure what is causing this, and I have a hard time reproducing it. I think it's a rare bug, and I'm not sure how this happens (my best guess is something like a race condition when the cell rerenders). It's not a bug usability problem, but if you happen to know the cause, it'll be great to patch it.

Aveline-art:
One other note. I found a couple of notes that might help streamline the hook:

[Element: mouseover event](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event)
[Determine which element the mouse pointer is on top of in JavaScript](https://stackoverflow.com/questions/8813051/determine-which-element-the-mouse-pointer-is-on-top-of-in-javascript)

app/frontend/src/components/Inputs/Calendar.tsx
gjagnoor: @Aveline-art I just fixed the bug. Can you confirm that you're also not able to reproduce it? I know it's a rare bug, but if it's not popping up after a few tries, especially after you leave the calendar component when the mouse is down and come back to the cells, then I think we should be fine. Thoughts?

Aveline-art:
@gjagnoor It looks like there are some merge conflicts on this. Can we resolve this so that a review can proceed?

app/frontend/src/components/Inputs/Calendar.tsx
```diff
@@ -25,33 +29,27 @@ interface CalendarRowProps extends React.PropsWithChildren {
 }
 
 interface CalendarCellProps {
-  columnNum: number;
-  onChange: (selected: boolean) => any;
-  rowNum: number;
-  selected?: boolean;
+  key: number;
+  data: string;
+  cell: any;
```
Aveline-art, line 34: Is it possible to limit the typing for `cell`? I see that below is is set as

```js
{
  row: number
  column: number
}
```

app/frontend/src/components/Inputs/Calendar.tsx
```diff
@@ -145,55 +145,58 @@ function CalendarRow(props: CalendarRowProps) {
...
+  }, [next]);
+
+  function mouseDown(e: any) {
+    e.preventDefault();
+    /* if (currentCell is filled) {
+      set is mouse down to true
+      set to select !next
+    } else {
+
+    } */
```
Aveline-art, line 174: Is this needed here?

app/frontend/src/components/Inputs/Calendar.tsx
```diff
@@ -145,55 +145,58 @@ function CalendarRow(props: CalendarRowProps) {
...
+  cell,
+  setToSelect,
+  setIsMouseDown,
+  isMouseDown,
+  data,
+  setData,
+  toSelect,
+}: CalendarCellProps) {
+  const [next, nextdata, handleSelect] = useDragToSelectUnselect(
+    { ...cell },
```
Aveline-art, line 158: I am not sure this destructuring does anything since it just recreates the cell object.

Aveline-art (changes requested):
@gjagnoor Left a couple of comments. The onclick functionality is pretty important but the rest are mostly stylistic.

app/frontend/src/components/Inputs/Calendar.tsx
```diff
@@ -145,55 +145,58 @@ function CalendarRow(props: CalendarRowProps) {
...
+        cell.row % 2 == 0 ? "dashed" : "solid",
+        next && "selected",
       )}
     >
       <div
         tabIndex={0}
         role="checkbox"
-        aria-checked={isSelected}
-        aria-label={`I am available on ${props.rowNum}, ${props.columnNum}`}
-        onClick={handleClick}
```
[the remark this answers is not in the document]
gjagnoor: Incorporated. Thanks for the feedback.

app/frontend/src/components/Inputs/Calendar.tsx
gjagnoor: No. Removed.

app/frontend/src/components/Inputs/Calendar.tsx
gjagnoor: It doesn't. I was testing something before and forgot to undo the change. Thanks for flagging.

app/frontend/src/components/Inputs/Calendar.tsx
gjagnoor: Done.

Aveline-art (approved):
Looks good!

I did discover a bug with the mouseup event not registering, but I'm having a hard time reproducing it consistently. I think it happens when the mouse moves rapidly (which might have caused a race condition with the JS stack). I'm not sure if most users would mouse the mouse really fast though, so I'm going to consider it not a concern until proven otherwise.

I also found that under certain conditions, the cells do end up being "highlighted and drag-able".

Screenshot 2023-02-21 at 8 50 33 PM

It seems this can be easily stopped by capturing the mousedown event object and [interrupt it's default function](https://stackoverflow.com/questions/5429827/how-can-i-prevent-text-element-selection-with-cursor-drag). I'm not too concerned about this. Feel free to merge this once this is fixed or open a new ticket to R&D a fix if it turns out to be more complicated.

Issue #220, opened 2022-04-13
Create Backend Schema in DRF


Aveline-art:
### Related Issue (Can be worked on in tandem)

- [x] #159 

Note: This issue can be done concurrently with its dependency.

### Overview

As part of our product we need to have a backend schema. For this issue we will add our tentative schema into our backend via DRF and a swagger UI.

### Action Items

- [x] See the db diagram below
- [x] Implement it using DRF
- [x] After implementation, add a swagger UI

### Resources/Instructions

[Resources](https://github.com/hackforla/CivicTechJobs/wiki/Resources)

Code for https://dbdiagram.io/d

```sql
// Creating tables
Table opportunities {
  id int [pk, increment] // auto-increment
  poster int
  job_title varchar
  project int
  max_commitment int [note: 'hours per week']
  min_commitment int [note: 'hours per week']
  max_duration int [note: 'days total']
  min_duration int [note: 'days total']
  positions_available int
  description varchar
  posted_date datetime [default: `now()`]
  role int
  experience int 
  required_tech int 
  preferred_tech int
}
```

Enzyme3:
defined the model in Django (with a few changes) and here's how it gets mapped. Can we confirm that this is the schema that should be used or if further changes should be made or reverted?
```
TABLE opportunity {
  id char(32) [pk] # changed from int to char(32) so that UUID could be used for primary key
  job_title varchar(100)
  poster varchar(100) # changed from int (foreign key) to varchar. will consist of email, and will not link to people depot
  description text
  project varchar(100) # changed from int (foreign key) to varchar. will consist of project name, and will not link to people depot
  max_commit_hours_per_week smallint # reworded var name to explicitly call out unit of measurement
  min_commit_hours_per_week smallint # reworded var name to explicitly call out unit of measurement
  num_positions_available smallint # reworded var name to explicitly call out unit of measurement 
  posted_date date # changed from datetime to date
  role varchar(2) # changed from int (foreign key) to varchar. will map to an enum defined in Django
  experience_in_months smallint
  required_exp varchar(100) # changed from int (foreign key) to varchar. will consist of project name, and will not link to people depot
  preferred_exp varchar(100) # changed from int (foreign key) to varchar. will consist of project name, and will not link to people depot
}
```

The enum for Roles are listed below. Are there any others that should be added?
```
BACKEND_DEVELOPER = 'BE'
FRONTEND_DEVELOPER = 'FE'
UX_DESIGNER = 'UX'
[...]

PR #330, opened 2022-10-23
Update schema to match VRMS

Enzyme3:
Update for #220

### Changes
- Updated recurring_event and project models to match VRMS (https://www.vrms.io/api/recurringevents) 
- Script to pull data from VRMS
- Added views/URLs for all models


Aveline-art:
@Enzyme3 It looks like there are some merge conflicts on this. Can we resolve this so that a review can proceed?

PR #338, opened 2022-12-02
lint ignore stage Dockerfile

Enzyme3:
### Changes
- Pin python alpine image to specific version

### DIFF
--- a/.github/workflows/build-deploy-stage.yml
+++ b/.github/workflows/build-deploy-stage.yml
@@ -29,7 +29,7 @@ permissions:
   contents: read
 
 jobs:
-  
+
   deploy:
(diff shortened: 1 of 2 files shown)


Aveline-art:
Just a quick question. Why pin curl to the specific version? I doubt curl will break based on how we are using it, but we do want to migrate to major python versions periodically to ensure repository freshness and patch core code bugs.

Enzyme3:
honestly, it's because the linter complains. If wanted, can add an exclusion rule for those two lines

Aveline-art:
@Enzyme3 Let's do an inclusion filter, and make sure that deployment does not happen if this fails (which it should, since docker failed).

Aveline-art (approved):
Looks good!

PR #356, opened 2023-02-28
Notification new

gjagnoor:
Resolves # 342

- new notification branch created so notification feature branch could be merged with `main` smoothly with as little hiccups as possible. 
- removed unnecessary files, cleaned up code

### DIFF
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -43,7 +43,10 @@ function NotificationBar({
         )}
       >
         <IconButton
-          addClass={`notification-bar-close appear-${autoHidden}`}
+          addClass={combineClasses(
+            "notification-bar-close",
+            autoHidden ? "hide" : "show"
+          )} // `notification-bar-close appear-${autoHidden}`
           iconUrl={iconX}
           label="close"
           onClick={() => setShowBar(false)}
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -1,4 +1,5 @@
 // Eternal Imports
+import { combineClasses } from "components/Utility/utils";
 import React, { useState, useRef } from "react";
 import { CSSTransition } from "react-transition-group";
 
@@ -7,10 +8,12 @@ import { iconX } from "../../assets/images/images";
 // Type declaration for props
 interface NotificationBarProps extends React.PropsWithChildren {
   autoHidden: boolean; // 1. Added new prop to indicate the bar disappearing on its own
+  color: String;
 }
 
 function NotificationBar({
   autoHidden = false,
+  color = "warning",
   ...props
 }: NotificationBarProps) {
   const [showBar, setShowBar] = useState(true);
@@ -29,7 +32,13 @@ function NotificationBar({
         if (autoHidden) setShowBar(false);
       }}
     >
-      <div ref={nodeRef} className="notification-bar">
+      <div
+        ref={nodeRef}
+        className={combineClasses(
+          "notification-bar",
+          `notification-bar-${color}`
+        )}
+      >
         <button
           onClick={() => setShowBar(false)}
           className="notification-bar-close"
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -9,11 +9,13 @@ import { iconX } from "../../assets/images/images";
 interface NotificationBarProps extends React.PropsWithChildren {
   autoHidden: boolean; // 1. Added new prop to indicate the bar disappearing on its own
   color: String;
+  text: String;
 }
 
 function NotificationBar({
   autoHidden = false,
   color = "warning",
+  text = "Content of the notification appears here",
   ...props
 }: NotificationBarProps) {
   const [showBar, setShowBar] = useState(true);
@@ -45,7 +47,7 @@ function NotificationBar({
         >
           <img src={iconX} height={"10px"} width={"10px"} alt="close" />
         </button>
-        <p className="text">{props.children}</p>
+        <div className="text">{text}</div>
       </div>
     </CSSTransition>
   );
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -28,7 +28,7 @@ function NotificationBar({
       in={showBar}
       nodeRef={nodeRef}
       classNames="notification-bar-fade"
-      timeout={300}
+      timeout={10000}
       appear
       unmountOnExit
       onEntered={() => {
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -36,7 +36,7 @@ function NotificationBar({
         >
           <img src={iconX} height={"10px"} width={"10px"} alt="close" />
         </button>
-        <p className="text-center">{props.children}</p>
+        <p className="text">{props.children}</p>
       </div>
     </CSSTransition>
   );
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -48,7 +48,9 @@ function NotificationBar({
           label="close"
           onClick={() => setShowBar(false)}
         />
-        <div className="notification-bar-text text-center">{text}</div>
+        <div className="notification-bar-text text-center">
+          {props.children}
+        </div>
       </div>
     </CSSTransition>
   );
--- a/app/frontend/src/components/NotificationBar/NotificationBar.tsx
+++ b/app/frontend/src/components/NotificationBar/NotificationBar.tsx
@@ -2,9 +2,10 @@
 import { combineClasses } from "components/Utility/utils";
 import React, { useState, useRef } from "react";
 import { CSSTransition } from "react-transition-group";
-
+import { IconButton } from "components/components";
 // Internal Imports
 import { iconX } from "../../assets/images/images";
+import { auto } from "@popperjs/core";
 // Type declaration for props
 interface NotificationBarProps extends React.PropsWithChildren {
   autoHidden: boolean; // 1. Added new prop to indicate the bar disappearing on its own
@@ -41,12 +42,12 @@ function NotificationBar({
           `notification-bar-${color}`
         )}
       >
-        <button
+        <IconButton
+          addClass={`notification-bar-close appear-${autoHidden}`}
+          iconUrl={iconX}
+          label="close"
           onClick={() => setShowBar(false)}
-          className="notification-bar-close"
-        >
-          <img src={iconX} height={"10px"} width={"10px"} alt="close" />
-        </button>
+        />
         <div className="notification-bar-text text-center">{text}</div>
       </div>
     </CSSTransition>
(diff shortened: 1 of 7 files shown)


Aveline-art (commented):
It looks like the design does not center the text or the close button. Is the intention behind this design that we be using custom classes to properly center them when used on pages? If we were to use the notification bar component on the search page, what is the plan to properly center it?

Screenshot 2023-03-01 at 3 24 06 PM

gjagnoor:
@Aveline-art Yes that was the intention - that when used on different pages the developer could customize the styles based on their needs. But now that I'm thinking again, I think it would make more sense to at least have a standard design if someone just wants to use it out of the box. I'll work on fixing the design.

gjagnoor:
Better? 

New Usage - 

```
<NotificationBar
  autoHidden={false}
  color={"success"}
  text={
    <p>
      This is a limited view. Please
      <button
        style={{
          border: "none",
          backgroundColor: "transparent",
          textDecoration: "underline",
          color: "blue"
        }}
      >
        attend onboarding
      </button>
      with Hack for LA or log in to see full opportunities.
    </p>
  }
/>
```

- Updated design to match closely to the figma design 
- included a color prop that the user can pass to the notification bar component. They have 4 options to choose from - theme options - "primary", "success", "warning", or "danger". Default is "warning"
Screen Shot 2023-03-08 at 6 57 41 PM


Screen Shot 2023-03-08 at 6 58 59 PM

app/frontend/src/components/NotificationBar/NotificationBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React, { useState, useRef } from "react";
+import { CSSTransition } from "react-transition-group";
+
+// Internal Imports
+import { iconX } from "../../assets/images/images";
+// Type declaration for props
+interface NotificationBarProps extends React.PropsWithChildren {
+  autoHidden: boolean; // 1. Added new prop to indicate the bar disappearing on its own
+  color: String;
+  text: String;
```
Aveline-art, line 12: Is this meant to be a `String` or `string` type?

app/frontend/src/components/NotificationBar/NotificationBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+function NotificationBar({
+  autoHidden = false,
+  color = "warning",
+  text = "Content of the notification appears here",
+  ...props
+}: NotificationBarProps) {
+  const [showBar, setShowBar] = useState(true);
+  const nodeRef = useRef(null);
+
+  // 2. Added CSS Transition component to wrap around notification bar
```
Aveline-art, line 24: It's not too important but we can probably clean up these comments. Thoughts?

app/frontend/src/components/NotificationBar/_NotificationBar_Close.scss
```diff
@@ -0,0 +1,97 @@
...
+@use "../Basics" as *;
+
+/*****************************
+*** Notification Bar class ***
+*****************************/
+$themes: "primary", "success", "warning", "danger";
+
+p {
+  padding: 0;
+  margin: 0;
```
Aveline-art, line 14: Let's avoid redefining whole tags for the entire document. This will get very messy down the line.

app/frontend/src/components/NotificationBar/_NotificationBar_Close.scss
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.notification-bar {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
```
Aveline-art, line 21: I think we can substitute these for classes in alignments and layout. It won't be a big improvement, though.

app/frontend/src/components/NotificationBar/_NotificationBar_Close.scss
```diff
@@ -0,0 +1,97 @@
...
+  padding: 1em;
```
Aveline-art, line 22: Let's use `px` whenever possible since our designs are based on px, but I also wonder if this can be removed?

app/frontend/src/components/NotificationBar/_NotificationBar_Close.scss
```diff
@@ -0,0 +1,97 @@
...
+.notification-bar-fade-exit {
+  opacity: 1;
+}
+
+.notification-bar-fade-exit-active {
+  opacity: 0;
+  transition-duration: 300ms;
+}
+
+.text {
```
Aveline-art, line 59: This name is too generic, so it increases the possibility of being overridden if we're careless in the future.

app/frontend/src/components/NotificationBar/_NotificationBar_Close.scss
```diff
@@ -0,0 +1,97 @@
...
+    "background-color": #72ca9b,
+    "color": black
+  ),
+  "warning": (
+    "background-color": $color-tan,
+    "color": black
+  ),
+  "danger": (
+    "background-color": #fa999c,
+    "color": black
```
Aveline-art, line 82: This is fine overall, but we don't really want to hardcode colors in case we do theming in the future. Let's set all these to colors we already have, like color-tan, and leave a comment that we will set these to different values in the future. Thoughts?

app/frontend/src/components/NotificationBar/index.tsx
```diff
@@ -0,0 +1 @@
+export { NotificationBar } from "./NotificationBar";
```
Aveline-art, line 1: I'm not sure what purpose this file serves?

app/frontend/src/index.scss
```diff
@@ -11,6 +11,7 @@
 @use "components/Dialog/Dialog";
 @use "components/Inputs" as *;
 @use "components/Navigation" as *;
+@use "components/NotificationBar/NotificationBar_Close";
```
Aveline-art, line 14: Let's rename this component to closer align with our naming scheme.

app/frontend/src/pages/Demo/Demo.tsx
```diff
@@ -14,9 +16,55 @@ function Demo() {
...
+        logoDesktop={logoHorizontal}
+        logoMobile={logoStacked}
+        menu={[
+          { name: "Hack for LA", link: "/" },
+          { name: "How to Join", link: "/qualifier" },
+          { name: "Projects", link: "/demo" }
+        ]}
+      />
+      <NotificationBar
+        text={
```
Aveline-art, line 32: I'm surprised typescript did not catch this, since these are not strings, theoretically. But if we plan to insert something more complicated that string into a `text` prop, using children would be the way to go. Children generally gives more freedom for the developer to insert complex content into a component.

app/frontend/src/components/NotificationBar/NotificationBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <CSSTransition
+      in={showBar}
+      nodeRef={nodeRef}
+      classNames="notification-bar-fade"
+      timeout={300}
```
Aveline-art, line 30: This fade value might be too low, since the element blinks out immediately, making it hard to read the notification.

app/frontend/src/pages/Demo/Demo.tsx
```diff
@@ -14,9 +16,55 @@ function Demo() {
...
+          <p>
+            This is a limited view. Please
+            <button
```
Aveline-art, line 35: Just trivia, but this should really be an `<a>` since it acts more like a link than a button.

Aveline-art (changes requested):
Everything is overall usable, but there's quite a lot that can be cleaned up. I think the most important ones to fix are the ones that might cause issues down the line.

One last thing I noticed: the height of the bar did not match the Figma specs.

Screenshot 2023-03-08 at 4 29 47 PM
Screenshot 2023-03-08 at 4 30 08 PM

gjagnoor:
@Aveline-art
Screen Shot 2023-03-14 at 2 45 08 PM 
- height of the notification bar does not match figma - fixed 
- text is not centered correctly - fixed 
- use <IconButton /> component to reduce redundancy - done 
- avoid redefining whole tags for the entire document - done 
- substitute use of flex box for classes in alignments and layout - done 
- use px for CSS units - done 
- avoid generic names - done 
- rename file according to convention - done
- use children instead of props for complicated insertions - done 
- let autohidden notification bar stay on screen longer - done  
- change button to a tag - done  
---
Other updates:
- added index.scss file. may need it further down the line, so thought maybe better to add now
- will follow figma very closely next time. Just getting used to it, so took me a little longer

gjagnoor:
changed the css logic a bit, but the previous comment still holds

https://github.com/hackforla/CivicTechJobs/pull/356#issuecomment-1468783416

app/frontend/src/components/NotificationBar/NotificationBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { CSSTransition } from "react-transition-group";
+import { IconButton } from "components/components";
+// Internal Imports
+import { iconX } from "../../assets/images/images";
+import { auto } from "@popperjs/core";
+// Type declaration for props
+interface NotificationBarProps extends React.PropsWithChildren {
+  autoHidden: boolean; // 1. Added new prop to indicate the bar disappearing on its own
+  color: String;
+  text: String;
```
Aveline-art, line 13: Pinging [this](https://github.com/hackforla/CivicTechJobs/pull/356/files#r1130236100) and [this](https://github.com/hackforla/CivicTechJobs/pull/356/files#r1130236718) in case it was missed last time. :)

app/frontend/src/components/NotificationBar/NotificationBar.tsx
```diff
@@ -0,0 +1,62 @@
+// Eternal Imports
+import { combineClasses } from "components/Utility/utils";
+import React, { useState, useRef } from "react";
+import { CSSTransition } from "react-transition-group";
+import { IconButton } from "components/components";
+// Internal Imports
+import { iconX } from "../../assets/images/images";
+import { auto } from "@popperjs/core";
```
Aveline-art, line 8: Can we take out this unused import?

app/frontend/src/components/NotificationBar/_NotificationBar.scss
```diff
@@ -0,0 +1,97 @@
...
+// Internal Imports
+@use "../Basics" as *;
+
+/*****************************
+*** Notification Bar class ***
+*****************************/
+$themes: "primary", "success", "warning", "danger";
+
+.notification-bar {
+  height: 55px;
```
Aveline-art, line 13: It looks like this was not used, which makes the component not match figma. Any specific reason why this was not used?

app/frontend/src/components/NotificationBar/_NotificationBar.scss
```diff
@@ -0,0 +1,97 @@
...
+  flex: 1;
+}
+
+.show {
+  visibility: visible;
+}
+
+.hide {
+  visibility: hidden;
+}
```
Aveline-art, line 26: We have [utility functions](https://github.com/hackforla/CivicTechJobs/blob/3c4ea0be150e02a25b6370e3a4d531026f79b78a/app/frontend/src/components/Basics/_utils.scss#L3) that can serve the function of these classes. Let me know if they are viable for your use cases.

There's also probably no need for both a show and hide, since if a class is removed, React will just remake the dom element, which resets everything to its default visibility.

Aveline-art (changes requested):
Looks good. Two things needs to be addressed before this matches the comp. in Figma:

1. The height of the component is slightly off.
2. I just noticed that in Figma, there is a variation that does not hide, and is not closable. I think the easiest way to do this is to expose another prop, maybe "closeable", that decides whether to render the close button. Thoughts?

I also noticed that because the icon button takes up dom space, the children in the comp is slightly off center horizontally. This is not quite noticeable, so there's no need to fix it. If you'd like to attempt that, however, you can attempt to use [float](https://developer.mozilla.org/en-US/docs/Web/CSS/float) to remove the icon from the dom space.

Aveline-art:
@gjagnoor Thank you for working on this. Since there has not been activity for a while, I'm going to close this and take over the ticket. Feel free to take up a different ticket when you have the time.

Issue #341, opened 2022-12-16
Component: Create the chevron scroll component


Aveline-art:
### Overview

As a developer we need to develop components that help us reuse code. For this issue, we will make the chevron scroll component.

### Action Items

- [x] Find component in figma under design system
- [x] Create it
- [x] Make sure it is accessible
- [x] Make sure it is reusable
- [x] For this component, make sure that it can accept any component within the scroll, including just text or links (the easiest might be to just have the inner items be children)
  - [x] Make sure the chevrons disappear when reaching the respective edge of the scroll box.

### Resources/Instructions

[Resources](https://hackforla.github.io/CivicTechJobs/resources/)

PR #378, opened 2023-04-06
Scroll

gjagnoor:
#341 

Please review this PR for functionality. I will clean up the code next. Please review the code in the next PR. 

scroll


gjagnoor:
@Aveline-art Is the Chip component you were working on ready to be used? I would like to use it for this component if that's ok with you. The chips you see in the gif attached are something I quickly put together to keep this feature work going. I'd rather replace these with your version to cut down on redundancy if we are ready to do so.

Aveline-art:
@gjagnoor The chip is merged already, so you can use it at your leisure. That said, it does not look like those are chip components, since they do not follow the rough specs in the design system.

Let me check in with the design team and get back on this.

Aveline-art (changes requested):
Functionally, it is missing two things:

- It's not scrollable by mouse wheel, for some reason. Looking at the code, I can't really tell why, since it's a div overflow at a glance.
- The scroll is too slow, so it's not usable at all. It also does not snap at the next children. Is there a reason why a custom scroll is used rather than something like `scroll-snap-stop`?

Aveline-art:
@gjagnoor Thank you for you work on this. As there has not been activity, I will close the ticket and throw it back into the backlog. Thank you!

PR #382, opened 2023-04-10
Add gap utilities 126

### LINKED ISSUE #126: Tweaking the layout components and utilities
### Overview
As a developer, we need to improve our design system by providing industry standards for layouts. For this issue we will add several spacing utilities to transform our layout classes.

### Action Items

- [x] Add gap utilities along with our margin and padding utilities.
- [x] Make sure that the gap utilities do not interfere with margin utilities or the standard 12 column design standard
- [x] If not already, move space utilities to the basics directory under a file named `_spacing.scss`
- [x] Also, add classes to layout that will shift its children to either to the start, center, or end (horizontally). This can probably be done through align-items, an attribute related to flex.

### Resources/Instructions
Please review our frontend architecture before you start.

Also note: it might be better to create a new layout class, maybe .grid-container that uses gaps rather than try to do recalculations based on both margins and gaps. This is fine since most DS uses either one or the other, but almost never both; and it is rare to find a use case for both.

bzzz-coding:
Fixes #126

### Changes

- Added gap utilities in app/frontend/src/components/Basics/_spacing.scss

### DIFF
--- a/app/frontend/src/components/Basics/_layout.scss
+++ b/app/frontend/src/components/Basics/_layout.scss
@@ -76,7 +76,7 @@ $columns: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
         // recalculates column size to accommodate gaps
         @if list.index("gap" "gap-x", $name) {
           // flex: 0 1 calc(8.32% * #{$col} -  (#{$size} * 8px) * (1 - #{$col} / 12));
-          flex: 0 1 calc(8.32% * $col -  ($size * 8px) * (1 - $col / 12));
+          flex: 0 1 calc(8.32% * $col - ($size * 8px) * (1 - $col / 12));
         }
       }
     }
--- a/app/frontend/src/components/Basics/_layout.scss
+++ b/app/frontend/src/components/Basics/_layout.scss
@@ -55,16 +55,16 @@ $columns: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
 
 @mixin col-size($col) {
   box-sizing: border-box;
-  flex: 0 1 calc(8.33% * #{$col});
+  flex: 0 1 calc(8.33% * $col);
 
   @each $name in map.keys($margin-abbrs) {
     @each $size in $spacing-sizes {
       &.#{$name}-#{$size} {
         // recalculates column size to accomodate margins
         @if list.index("mx" "my" "px" "py", $name) {
-          flex: 0 1 calc(8.33% * #{$col} - #{$size} * 16px);
+          flex: 0 1 calc(8.33% * $col - $size * 16px);
         } @else {
-          flex: 0 1 calc(8.33% * #{$col} - #{$size} * 8px);
+          flex: 0 1 calc(8.33% * $col - $size * 8px);
         }
       }
     }
@@ -75,7 +75,6 @@ $columns: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
       .#{$name}-#{$size} > & {
         // recalculates column size to accommodate gaps
         @if list.index("gap" "gap-x", $name) {
-          // flex: 0 1 calc(8.33% * #{$col} -  (#{$size} * 8px) * (1 - #{$col} / 12));
           flex: 0 1 calc(8.33% * $col - ($size * 8px) * (1 - $col / 12));
         }
       }
--- a/app/frontend/src/components/Basics/_layout.scss
+++ b/app/frontend/src/components/Basics/_layout.scss
@@ -75,8 +75,8 @@ $columns: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
       .#{$name}-#{$size} > & {
         // recalculates column size to accommodate gaps
         @if list.index("gap" "gap-x", $name) {
-          // flex: 0 1 calc(8.32% * #{$col} -  (#{$size} * 8px) * (1 - #{$col} / 12));
-          flex: 0 1 calc(8.32% * $col - ($size * 8px) * (1 - $col / 12));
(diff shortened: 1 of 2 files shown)


MattPereira:
I'm still getting acquainted with Sass so I'll leave the approval up to Ava but I did do some manual testing of the new gap utility and it seems to be functioning as intended

<details><summary>See gap-0</summary>

image

</details>


<details><summary>See gap-2</summary>

image

</details>

<details><summary>See gap-x-2</summary>

image

</details>

<details><summary>See gap-y-2</summary>

image

</details>

Aveline-art (commented):
Code looks good overall, but I can't seem to get it to work with our 12 column layout. Any usage notes on how to use the gap utilities with our layout?

Screenshot 2023-04-12 at 7 56 45 PM

For reference, this is the code I am using:

```jsx
{
  [
    ["", "red"],
    ["ml-3", "blue"],
    ["pl-3", "orange"],
    ["gap-x-3", "MediumVioletRed"],
  ].map(([space, color], index) => {
    return (
      <div
        key={index}
        className={`flex-container mb-3 ${[0, 3].includes(index) ? space : ""}`}
        style={{ height: "50px" }}
      >
        {Array(4)
          .fill(3)
          .map((txt, jindex) => {
            return (
              <div
                key={jindex}
                className={`col-3 ${[1, 2].includes(index) ? space : ""}`}
                style={{
                  border: `${color} solid 1px`,
                }}
              >
                {txt}
              </div>
            );
          })}
      </div>
    );
  });
}
```

bzzz-coding:
Ah it's because gaps are included in the total width of the flexbox, so if each .col-3 takes up 25% of the flexbox width, once we add gaps, the total width would exceed 100%, resulting in flex items being pushed to the next row. 

To fix this, I edited the _layout.scss file to adjust column widths to accommodate gaps. 

However, I have a question regarding line 78—both line 78 and line 79 seem to do the same job. I wrote line 79 but included line 78 as a comment for now in case that's the preferable syntax, as it's consistent with the existing code in this file such as line 65, but I'm not familiar with this syntax (specifically, Sass string interpolation used in calculation). Is there a reason that we might want to use one or the other?

Aveline-art:
@bzzz-coding Great question. I think I must have missed it when it was being written. I think overall having cleaner code is a win, so feel free to put whichever format you feel is best as long as it achieves the same thing.

app/frontend/src/components/Basics/_layout.scss
```diff
@@ -69,6 +69,18 @@ $columns: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
...
     }
   }
+
+  @each $name in map.keys($gap-abbrs) {
+    @each $size in $spacing-sizes {
+      .#{$name}-#{$size} > & {
+        // recalculates column size to accommodate gaps
+        @if list.index("gap" "gap-x", $name) {
+          // flex: 0 1 calc(8.33% * #{$col} -  (#{$size} * 8px) * (1 - #{$col} / 12));
+          flex: 0 1 calc(8.33% * $col - ($size * 8px) * (1 - $col / 12));
```
Aveline-art, line 79: Hmm, this works, but it does tend to take away more space from smaller columns than bigger ones, leading to big columns appearing bigger and smaller columns smaller.

It's not a big deal, I think, since I suspect we'll use gaps only in very rare cases, and it might be a big lift to remove width via column ratios.

app/frontend/src/components/Basics/_spacing.scss
```diff
@@ -23,20 +23,25 @@ $padding-abbrs: (
   "px": padding-right padding-left,
   "py": padding-top padding-bottom,
 );
-$spacing-abbrs: map.merge($margin-abbrs, $padding-abbrs);
-$spacing-sizes: 0, 1, 2, 3, 4, 5, 6;
+$gap-abbrs: (
+  "gap": gap,
```
Aveline-art, line 27: I'm thinking these will be easier to use if they're named "g, gx, gy". Thoughts?

app/frontend/src/components/Basics/_layout.scss
bzzz-coding: Yes, I did this math assuming all columns have the same width. The tricky thing about trimming widths proportionately is that I don't know how to get the number of columns/gaps in the flexbox in Sass/CSS. I feel like it might involve JS, but I need to research how to run a script and import values to this .scss file or generate SASS in a JS file. This may take a while, so I'll merge this PR now.

Alternatively, we could use `grid `instead of `flexbox` with gaps. Grid will refer to its `grid-template-column` and maintain the ratio of columns when applying gaps in between. Example: https://codepen.io/bzzz-coding/pen/WNawreW

app/frontend/src/components/Basics/_spacing.scss
bzzz-coding: I always wondered why margins and paddings have short utility class names like `p-3`, `pt-5`, `mx-2`, `mb-4`, while gap utilities use the word 'gap' (maybe because it's already short to begin with?). Personally, I have no preference but CSS frameworks like Bootstrap and Tailwind both use gap class names like `gap-1`, `gap-x-2` and `gap-y-3` so I thought maybe that's the convention/industry standard 🤔

app/frontend/src/components/Basics/_layout.scss
Aveline-art: Sounds good.

app/frontend/src/components/Basics/_spacing.scss
Aveline-art: Hmm, I know the 12 column system is a de-facto standard, but I've never heard of the class names being a standard. Curious to know more :hmm:

bzzz-coding: Maybe there's no standard? I have no idea. I was referencing [bootstrap](https://getbootstrap.com/docs/5.0/utilities/spacing/#gap) and [tailwindCSS](https://tailwindcss.com/docs/gap).

PR #386, opened 2023-04-18
Created Notification Component and Draft Transition Wrapper Component

### LINKED ISSUE #342: Component: Create the notification bar component
### Overview

As a developer we need to develop components that help us reuse code. For this issue, we will make the chevron scroll component.

### Action Items

- [ ] Find notification bar in figma under design system
- [ ] Create it
- [ ] Make sure it is accessible
- [ ] Make sure it is reusable
- [ ] This component is rather simple, but one optional improvement would be to have different modes:
  - [ ] Stay on the page mode.
  - [ ] Disappear after a set period mode (this should accept any animations created via [react-transition](https://reactcommunity.org/react-transition-group/))

### Resources/Instructions

Aveline-art:
Fixes #342 

### Changes

- Created `<Notification>` component with three main props:
  - closable: adds a x button to close the notification
  - fade: adds a fade transition effect on enter and exit
  - autohidden: automatically dismisses the notification over time 
- Created `<Wrapper>` component in transition to aide in wrapping transitions, which contains a lot of boilerplate to properly handle its props to produce predictable effects
  - I will open a new ticket to refactor the dialog component with the `<Wrapper>`. We might also need some discussion on how the `<Wrapper>` should work as a component.
- Added jest tests for Notification. This is the first test in typescript; which is very much the same as in js. I'll write a new ticket for the code migration for existing tests. 
  - It looks like jsdom needs to be loaded with stylesheets for react-testing to be able to use `toBeVisible`. Since our stylesheet is compiled, it'll be hard to do that, so we are testing for `.hidden` class as a proxy. If we use proper ARIA, we could also test for ARIA attributes.
- Updated python deps as per dependabot alert 
  - https://github.com/hackforla/CivicTechJobs/security/dependabot/28
  -  https://github.com/hackforla/CivicTechJobs/security/dependabot/27


MattPereira (commented):
Really appreciate the opportunity to learn by attempting to review this one :sweat_smile: 

I am curious about the speed at which the `autoHidden` prop hides the `Notification`. Apologies if I'm misinterpreting, but the speed appears to be controlled by the `timeout={400}` prop passed into the `CSSTransition` component that is returned by the `TransitionWrapper` component. Is there a prop you can pass to the `Notification` component directly to control the auto-hide speed or is the intention to have it always hide in less than a half second?

app/tests/components/Notification.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  test("Notification component closable", async () => {
+    const user = userEvent.setup();
+    render(
+      <Notification show={true} closable>
+        This is a warning
+      </Notification>
+    );
+    await user.click(screen.queryByRole("button")!);
+  });
```
MattPereira, line 33: Should there be expectations for this test?

Aveline-art: Whoops! Thanks for catching that. Looks like we'll need tests for our tests 😉

Aveline-art:
@MattPereira The documentation is a bit unclear on what the timeout does for CSSTransition. From what I gather, it exists fairly independently of the transition time in CSS. I believe what it does is control when the CSS classes are added and removed. I think it gives an "out" of sorts to the component, since there's no way, otherwise, of it knowing when a transition has fully completed its animation.

That said, I do think it should work like the base Transition element, in that it applies or overrides the transition time specified in SCSS, so that there would not be a need to redundantly replicate this value. I wonder why they need to be separate.

bzzz-coding (approved):
Hey @Aveline-art , thanks for working on this issue and requesting a code review! It was great I got to learn about React Transition Group and it was fun working through these changes with @MattPereira . 
I think I have a good understanding of 80% of the code and the demo worked. I need time to learn more about Jest but don't want to delay the merge of this PR so I'll approve it now. 

Thanks again! ❤️

PR #389, opened 2023-04-25
Add pre-commit hook for prettier

### LINKED ISSUE #147: Add prettier as a pre-commit hook
### Dependency

_No response_

### Overview

As a developer, we want to make sure the process of developing is easy for us, so that we can deliver high-quality work at a quickened pace. For this issue we will install prettier as a pre-commit hook.

### Action Items

- [ ] Configure the repo such that prettier runs on commit.

### Resources/Instructions

[Resources](https://github.com/hackforla/CivicTechJobs/wiki/Resources)
[Pre-commit Hook](https://prettier.io/docs/en/precommit.html)

Considerations: Since we use docker as our container, prettier, even if added to package.json, will not be installed globally or locally for our project. In order to use prettier as a pre-commit hook, one part of this issue is finding a way to use prettier while it exists inside a docker container. One solution to explore might be to customize our setup such that when a commit is made, a command runs prettier installed inside the container. However, feel free to explore other options as well. The ideal solution for us should be simple, and easy for the whole team to implement. Ideally, it should be baked into our repository.

MattPereira:
Fixes #147

### Changes
- Added prettier hook to the `.pre-commit-config.yaml` file as recommended by [option 3 from the Prettier documentation](https://prettier.io/docs/en/precommit.html#option-3-pre-commithttpsgithubcompre-commitpre-commit) because I think it best fits our use case of working with a multi-language project
- Chose `v2.7.1` as it seems to be the latest stable version of prettier supported by the [mirrors-prettier repo](https://github.com/pre-commit/mirrors-prettier) that the Prettier docs recommend using
- Added `nodejs` and `npm` to apk dependencies in `linter.dockerfile` because prettier requires them or else `docker compose build linter` will fail

### Testing
-  Rebuild linter image with `docker compose build linter`
- Disable prettier on save if you have that configured and then mess up the spacing on one of the files.
- After adding changes to staging area with `git add .` try running `docker compose run linter` and it should reformat the staged file with output in terminal that looks like this

image

### DIFF
--- a/dev/linter.dockerfile
+++ b/dev/linter.dockerfile
@@ -13,8 +13,8 @@ RUN apk add --no-cache \
   python3-dev~=3.9 \
   musl-dev~=1.2 \
   py3-pip~=20.3 \
-  nodejs \
-  npm
+  nodejs~=16.20 \
+  npm~=8.1.3
 
 
 # install python dependencies
--- a/dev/linter.dockerfile
+++ b/dev/linter.dockerfile
@@ -12,7 +12,9 @@ RUN apk add --no-cache \
   python3~=3.9 \
   python3-dev~=3.9 \
   musl-dev~=1.2 \
-  py3-pip~=20.3
+  py3-pip~=20.3 \
+  nodejs \
+  npm
 
 
 # install python dependencies
@@ -21,9 +23,10 @@ ENV PYTHONUNBUFFERED=1
 RUN ln -sf python3 /usr/bin/python &&\
   python -m ensurepip &&\
   pip3 install --no-cache --no-cache-dir --ignore-installed --upgrade \
-    pip==22.1 \
-    pre-commit==2.19 \
-    setuptools==62.3
+  pip==22.1 \
+  pre-commit==2.19 \
+  setuptools==62.3
+
 
 WORKDIR /src
 ENTRYPOINT ["pre-commit"]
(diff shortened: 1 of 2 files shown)


dev/linter.dockerfile
```diff
@@ -12,7 +12,9 @@ RUN apk add --no-cache \
   python3~=3.9 \
   python3-dev~=3.9 \
   musl-dev~=1.2 \
-  py3-pip~=20.3
+  py3-pip~=20.3 \
+  nodejs~=16.20 \
+  npm~=8.1.3
```
Aveline-art, line 17: Any reason why this was selected over other versions?

Aveline-art (approved):
Looks good! Thanks for doing this.

dev/linter.dockerfile
MattPereira: I figured out those versions by building the image without specifying a version and then running the container and inspecting what versions were installed :joy: 

I am interested to learn if you have an approach or process for determining which version?

Aveline-art: I was actually curious why older versions were used over the latest. It makes sense to use the versions that work/come with the default alpine image.

I wonder if it will also make sense to add a command to update the apt package repositories and download the latest node/npm versions. That said I'm not particularly worried if the linter container breaks, since it's non-essential, so this is by no means pressing. Thoughts?

PR #393, opened 2023-05-04
Bash script for aliases

### LINKED ISSUE #192: Create bash script to run shortened version of useful commands
### Overview

As a developer, we need to make development friendly so that less time is spent on debugging and more time on building a valuable product. For this issue we will leverage bash scripting to quickly run our cumbersome commands.

### Action Items

- [ ] Develop a script (or series of scripts), that can be customized to run shortened commands (see below for more info)

### Resources/Instructions

This issue is inspired by how npm leverages package.json to quickly run certain commands. For those who do not know, package.json contains a script section which shortens certain commands. For example, let's say we have these scripts:

```json
"scripts": {
  "watch": "webpack --mode development --watch",
  "build": "webpack --mode production --output-path ./frontend/static/frontend"
},
```

Rather than type in `webpack --mode production --output-path ./frontend/static/frontend` in our commandline, all we need to to type is `npm run build`, which is shorter and easier to recall (as well as more intuitive). However, since we do not utilize npm, but rather docker, we must depend on shell scripting to achieve the same effect.

For this issue, this would be the ideal way for the script to function:

The user runs the script with some arguments, for example: ./script.sh $arg
```bash
Read a config file containing a list of shortened-ver./commands:
  If the shortened-ver. matches $arg
    Run the associated command
    exit
  Else:
    echo a message saying to run a flag, '-h' to get more info.
    exit

Likewise, if $arg is the '-h' flag:
  echo the help message
  exit
```

[Resources](https://github.com/hackforla/CivicTechJobs/wiki/Resources)
[Bash scripting cheatsheet](https://devhints.io/bash)

MattPereira:
Fixes #192

### Changes

- Added shell script `alias.sh` that leverages the `alias-config.txt` file allowing us to create shortened versions of longer shell commands and run them like `./alias.sh test`
- Populated `alias-config.txt` with as many commands as I could find from our mkdocs

### Considerations 
- You will most likely need to give the script file permission to execute with a command like `chmod u+x alias.sh` if you want to run script with `./alias.sh start` , but you might be okay to use `bash ./alias.sh start` without modifying permissions
- While researching, I was warned that not all operating systems ship with the `script` command installed by default.
     -  The trade off is that if we use `eval "$command"` instead then we will have to add `-T` option to `docker run` commands in the `alias-config.txt` file because they require a `TTY` context by default
     - Also when running `./alias.sh test` the terminal won't show the pretty colors if we decide to use `eval` instead of `script`

### DIFF
--- a/alias-config.txt
+++ b/alias-config.txt
@@ -1,0 +1,6 @@
+build: docker compose build
+start:docker compose up
+stop:docker compose down
+build-linter:docker compose build linter
+lint:docker compose run linter
+test:docker compose run webpack npm run test
--- a/alias-config.txt
+++ b/alias-config.txt
@@ -2,5 +2,5 @@ build: docker compose build
 start:docker compose up
 stop:docker compose down
 build-linter:docker compose build linter
-lint:docker compose run linter
-test:docker compose run webpack npm run test
+lint:docker compose up linter
+test:docker compose run -T webpack npm run test
--- a/alias.sh
+++ b/alias.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Configuration file containing shortened versions and associated commands
+config_file='alias-config.txt'
+
+# Check if the config file exists
+if [ ! -f "$config_file" ]; then
+  echo "Configuration file not found: $config_file"
+  exit 1
+fi
+
+# Function to display help message
+show_help() {
+  echo "Usage: $0 <alias>"
+  echo ""
+  echo "Options:"
+  echo "  -h   Show help message"
+  echo ""
+  echo "Available commands:"
+  while IFS=':' read -r alias command || [ -n "$alias" ]; do
+    printf "  %-8s %s\n" "$alias" "$command"
+  done < "$config_file"
+}
+
+# If no arguments provided or the argument is -h, show help message
+if [ $# -eq 0 ] || [ "$1" = "-h" ]; then
+  show_help
+  exit 0
+fi
+
+# Read the config file to find the matching shortened version
+while IFS=':' read -r alias command; do
+  if [ "$alias" = "$1" ]; then
+    # The -q flag suppresses the output from the script command itself
+    # The -c flag specifies the command to run and /dev/null discards the typescript file that script would normally generate
+    script -q -c "$command" /dev/null
+    exit 0
+  fi
+done < "$config_file"
+
+# If no matching shortened version is found, display an error message
+echo "Unknown alias (shortcut name for command): $1"
+echo "Run $0 -h for more information."
+exit 1
--- a/alias.sh
+++ b/alias.sh
@@ -3,12 +3,6 @@
 # Configuration file containing shortened versions and associated commands
 config_file='alias-config.txt'
 
-# Check if the config file exists
-if [ ! -f "$config_file" ]; then
-  echo "Configuration file not found: $config_file"
-  exit 1
-fi
-
 # Function to display help message
 show_help() {
   echo "Usage: $0 <alias>"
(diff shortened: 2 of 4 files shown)


alias.sh
```diff
@@ -0,0 +1,44 @@
...
+  show_help
+  exit 0
+fi
+
+# Read the config file to find the matching shortened version
+while IFS=':' read -r alias command; do
+  if [ "$alias" = "$1" ]; then
+    # The -q flag suppresses the output from the script command itself
+    # The -c flag specifies the command to run and /dev/null discards the typescript file that script would normally generate
+    script -q -c "$command" /dev/null
```
Aveline-art, line 36: I'm not exactly sure for the usecase of this. Is there a reason why we are not going directly to $command?

`script` also does not have the -c flag on my machine, so this does not seem to work out of the box for some developers. Can we make this script more friendly for more machines? That said, I don't think we should support windows, as it looks like no one in the team use windows. Thoughts?

alias-config.txt
```diff
@@ -0,0 +1,6 @@
+build: docker compose build
+start:docker compose up
+stop:docker compose down
+build-linter:docker compose build linter
+lint:docker compose run linter
+test:docker compose run webpack npm run test
```
Aveline-art, line 6: This is usable but not very extensible. For example, if I want to do `docker compose build --no-cache`, I would need to create a new command, maybe build-no-cache, which starts to defeat the purpose of the script.

I think it is a good improvement to add kwargs or args support so we can run build commands with args and flags. Thoughts?

```
sh alias.sh build
sh alias.sh build --no-cache
sh alias.sh build linter
sh alias.sh build linter --no-cache
```

Aveline-art (commented):
Just a couple of thoughts.

One more thing I would add is we can probably document information on how to use this on our docs site.

alias.sh
MattPereira: If you "go directly to $command" the script will exit with `the input device is not a TTY` 

You can get around the TTY by passing `-T` option to "allocate a psuedo-TTY" but then the output in terminal looks a little different. So like for `./alias.sh test` it would look like this instead of having the nice default colors:

image

</details>

But perhaps I'm too attached to the colors and psuedo-TTY output is fine?

Aveline-art: I'm a bit confused. Are we seeing the same thing? What is the program you are using to run your clis?

Screenshot 2023-05-06 at 3 05 00 PM
Screenshot 2023-05-06 at 3 05 23 PM

Aveline-art: Ahh, I see this is happening for the test and lint commands. Interesting.

Aveline-art: That said it looks like this configuration works out of the box with the behavior that we would expect it to. Any thoughts on whether this is a viable path forward? 

Screenshot 2023-05-06 at 3 14 59 PM

MattPereira: > What is the program you are using to run your clis?

Ubuntu runs `.sh` files with with the `bash` executable found in `/usr/bin/bash` which I beleive is the same for all Unix-based systems including macOS

> Any thoughts on whether this is a viable path forward?

Looks like swapping `compose run` for `compose up` is viable for the `lint` but not the `test` command.

image

However, we can still run the `test` command with `test:docker compose run -T webpack npm run` 

image

alias-config.txt
MattPereira: I think allowing for additional args and flags also defeats the purpose of the script assuming that the intention is to offer devs a more efficient method for executing longer commands without having to remember all of the necessary arguments and/or flags

I'm curious if @bzzz-coding has a preference or alternative perspective to offer?

PR #397, opened 2023-05-13
Corrected links and added debug to aws env

Aveline-art:
Partial fix for #78

### Changes

- This is a partial fix for #78
- Since our staging website is already up, I have decided to do a quick PR to fix a lot of unfortunate bugs such as:
  - urls leading to the wrong page  
  - security vulnerabilities with our `<a>` tags
  - the staging env being hosted in debug mode

### DIFF
--- a/app/frontend/src/components/Buttons/Button.tsx
+++ b/app/frontend/src/components/Buttons/Button.tsx
@@ -47,6 +47,7 @@ function Button({
       )}
       href={props.href}
       target={props.href ? props.target : undefined}
+      rel={props.href ? "noopener noreferrer" : undefined}
       onClick={props.onClick}
     >
       {props.children}
--- a/app/frontend/src/components/Buttons/Button.tsx
+++ b/app/frontend/src/components/Buttons/Button.tsx
@@ -16,12 +16,14 @@ interface ButtonAnchorProps extends ButtonSharedProps {
   href: string;
   onClick?: never;
   target?: "_blank" | "_self" | "_parent" | "_top";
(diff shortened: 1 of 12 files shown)


Aveline-art:
@MattPereira @bzzz-coding This is a somewhat emergency PR to patch up some security vulnerabilities. I plan to merge this in tomorrow without or w/o a pr. If you have time, I recommend looking it over and giving some thoughts. This PR changes a number of files, so it has a chance of causing merge conflicts.

app/frontend/src/components/Buttons/Button.tsx
```diff
@@ -47,6 +47,7 @@ function Button({
       )}
       href={props.href}
       target={props.href ? props.target : undefined}
+      rel={props.href ? "noopener noreferrer" : undefined}
```
bzzz-coding, line 50: @Aveline-art could you help me understand why we want to check if there is an href first, instead of just setting rel to "noopener noreferrer" all the time?

bzzz-coding (approved):
@Aveline-art Thanks for fixing the links and making the site more secure for visitors! I pulled your code and tested on my end and everything seems to work as intended. ✨🚀

app/frontend/src/components/Buttons/Button.tsx
Aveline-art: The check is needed because the `rel` attribute only exists for `<a>` and not `<button>` tags. I think, technically, adding it to buttons would not cause any issues in rendering, however.

PR #426, opened 2023-08-09
Installed and set up tailwindCSS

### LINKED ISSUE #424: Install and set up TailwindCSS
### Overview
Install and set up TailwindCSS for efficiently adding styles to components. 

### Action Items
- Install TailwindCSS
- Configure tailwind and webpack
- Build Docker image
- Document the setup and configurations

### Resources/Instructions
-Setup Tailwind CSS in a React project configured from scratch with Webpack https://dev.to/ivadyhabimana/setup-tailwind-css-in-a-react-project-configured-from-scratch-a-step-by-step-guide-2jc8

bzzz-coding:
Fixes #424

### Changes

- Installed TailwindCSS
- Added DemoTailwind page for testing
- Added `/demo-tailwind` path to see tailwind demo components

<details>
  <summary>DemoTailwind Page @/demo-tailwind</summary>
  <br>
  
  <br>
</details>


MattPereira:
image

frontend/tailwind.config.js
```diff
@@ -0,0 +1,16 @@
...
+//   plugins: [],
+// }
+
+module.exports = {
+  content: ["./src/pages/Demo/DemoTailwind.tsx"], // "./src/**/*.{js,jsx,tsx}", "./templates/index.html"
+  theme: {
+    extend: {},
+  },
+  plugins: [],
+};
```
MattPereira, line 16: my suggestion is to disable the tailwind normalization's completely like this

```javascript
module.exports = {
  corePlugins: {
    preflight: false,
  },
  // ...
}
```

and then we hope that the Sass normalizations cover bases as we build new pages with tailwind :crossed_fingers:

bzzz-coding: Hey @MattPereira thanks for the suggestion✨! I tried disabling the preflight but it will affect components that rely only on tailwind for styling. However, it is a quick way for devs to check whether certain style changes are caused by tailwind preflight. I added these lines to tailwind.config.js as comments for future reference.

bzzz-coding:
> @bzzz-coding i think we might be okay relying on the existing Sass normalizations for now. I share your concern about not having the tailwind normalization as we build new pages with tailwind but maybe the Sass will cover it? What do you think?
> 
> The only other alternative I can think of is trying to "merge" the normalizations between Sass and Tailwind by selectively overriding styles but that sounds like a total nightmare

I identified a few major differences between Tailwind preflight and our current normalize.scss by examining our demo page:
- default margins for headings: TW removes all default margins for headings
- img display in buttons: TW sets all img to have a display: block; 

So I 
- added a rule in _Button.scss to set img and SVG within a button to have a display: inline
- added margin for h1 and h2 in _typography.scss

Now at least our existing components and pages look similar to before installing Tailwind, and new components that don't rely on our Sass styles can be styled with Tailwind base and utility classes.

bzzz-coding:
Linter kept complaining about the @tailwind directives in the styles.css file I added as a step of installing Tailwind, so I removed the styles.css file and directly imported "tailwindcss/tailwind.css" which performs the same way according to [this documentation](https://v2.tailwindcss.com/docs/installation): 
`If you’re working in a JavaScript framework like React or Vue that supports directly importing CSS files into your JS, you can also skip creating a CSS file altogether and import tailwindcss/tailwind.css instead which has all of these directives already in place:`

Screenshot 2023-08-09 at 2 41 27 PM

MattPereira:
@bzzz-coding love the solution you came up with! ( much better idea than trying to use tailwind without the preflight :sweat_smile:  )


The only remaining TW preflight conflict I've found is the hr that shows up on the qualifier page

<details><summary>See hr w/o margin when TW preflight is active</summary>

Screenshot from 2023-08-09 15-08-49

</details>

<details><summary>See hr w/ margin on stage.civictechjobs.org</summary>

Screenshot from 2023-08-09 15-08-22


</details>

frontend/src/components/Basics/_typography.scss
```diff
@@ -7,9 +7,11 @@
   @if $size == 1 {
     font-size: 3rem;
     line-height: 3.625rem;
+    margin: 0.67em 0; // Temporary: added margin for current components
   } @else if $size == 2 {
     font-size: 2.25rem;
     line-height: 2.75rem;
+    margin: 0.83em 0; // Temporary: added margin for current components
```
MattPereira, line 14: So the TW preflight styles override all of the styles in `_normalize.scss` right ?

But moving the styles into this file overrides the TW preflight?

frontend/tailwind.config.js
MattPereira: lol turned out to be a not good suggestion but I do appreciate the sentiment :sweat_smile:

frontend/src/components/Basics/_typography.scss
bzzz-coding: I believe so. It's not ideal--I think the standard is to remove all styles for headings in the normalized css, but I didn't want our demo page to look too different from before so I added these margins only for h1 and h2 in _typography.scss, but by doing so, new components/pages will also have h1 and h2 with these margins. My hope is that in the near future we can update component styles so we can 100% rely on TW preflight. What are your thoughts on this?

MattPereira (approved):
@bzzz-coding super impressive work on this one!!! :raised_hands: 

love the elegant sidestep of the grumpy linter

apologies for initially suggesting that most of the heavy lilfting would involve webpack when the real enemy turned out to be a normalization specificity battle

Thanks for figuring this out so we can build all the new pages and components with TW :partying_face:

bzzz-coding:
> @bzzz-coding love the solution you came up with! ( much better idea than trying to use tailwind without the preflight 😅 )
> 
> The only remaining TW preflight conflict I've found is the hr that shows up on the qualifier page
> 
> See hr w/o margin when TW preflight is active
> Screenshot from 2023-08-09 15-08-49
> 
> See hr w/ margin on stage.civictechjobs.org
> Screenshot from 2023-08-09 15-08-22

Ah great job catching it! I didn't notice this one. I'm curious about what you think but I actually prefer the TW version. It does look different from the staging site but maybe we can add TW margin utility classes to achieve the same spacing instead of having the browser default margins? I personally think it's good practice to remove all default margins for hr but I'm happy to make changes too!

frontend/package-lock.json
bzzz-coding, line 1: @MattPereira Hey Matt did you notice that for several dependencies the `"dev": true,` was removed? I didn't intentionally do this but maybe it's because I had to build the docker image again and run npm install...

frontend/src/components/Basics/_typography.scss
MattPereira: I think its a question of tailwind philosophy that we should discuss. My understanding is that with tailwind you are supposed to apply absolutely every style with classNames right? 

so if we wanted a global h2 style the only option that adheres to tailwind best practices is to create a component just for h2 styles that we would then import anytime we wanted use an h2 element?

or we could just run an extra global stylesheet in addition to tailwind. which do you prefer?

MattPereira: Maybe it would be better to collect any normalization conflict fixes in the `index.scss` file ?

frontend/package-lock.json
MattPereira: nah I hadn't noticed until now. that is weird, but its also strange how `react` and `react-router-dom`are in the `devDependencies` instead of regular dependencies for the `package.json`

image


</details>

I just googled it and turns out the separation is only semantically meaningful so it should be nbd that `"dev": true,` was removed for a bunch of the underlying package dependencies

bzzz-coding: This is so strange. Any idea how these changes could have happened? Also, the ones in "dependencies" seem like they should be "devDependencies".

frontend/src/components/Basics/_typography.scss
bzzz-coding: That makes sense. I'll move the styles.

frontend/package-lock.json
bzzz-coding: I reinstalled some dev dependencies (tailwindcss, postcss, postcss-loader, autoprefixer) and now package.json and package-lock.json look more reasonable.

PR #442, opened 2023-08-30
Authentication page layout

### LINKED ISSUE #440: build login/signup page layout
### Overview
We want to begin work on the Login in/ Sign up pages. For this issue, we will setup the new component layout and route.

### Action Items
- [x] Create page component for authentication page
- [x] Add "/login" and "/signup" paths to router
- [x] Enable tailwind css for the authentication page component
- [x] Update navbar so it looks like figma when user is visiting login or signup page

### Resources/Instructions
- [Figma Designs](https://www.figma.com/file/G5bOqhud6azbxyR9El9Ygp/Civic-Tech-Jobs?type=design&node-id=1-26&mode=design&t=Py7eAJ2nlHoaLAfL-0)

MattPereira:
Fixes #440

### Changes
- Opted for multiple paths so that users can be sent a direct link to the sign up page
- Converted `HeaderNav` to all tailwind and only using horizontal CTJ logo since no stacked logo seen in figma
- Created a `InputGroup` component with tailwind because not sure about using or converting existing `Textfield.tsx` 

### Discussion
- The `HeaderNav` is using `a` element links instead of react router `Link` components because the `HeaderNav` component is outside the context of react router that is created in `Router.tsx`
- How to enable "hot reload" for this project? Clicking the refresh button in browser to see each css change is a nightmare
- Is there a better way to structure or organize the components related to the authentication page?

### Follow Up Issues
- #443
- #444
- #445

<details><summary>See Desktop</summary>

image

image

</details>

<details><summary>See Mobile</summary>

image

image

</details>


bzzz-coding:
- The HeaderNav is using a element links instead of react router Link components because the HeaderNav component is outside the context of react router that is created in Router.tsx
I wonder if we should reorganized the router and take advantage of [nested routes](https://reactrouter.com/en/main/start/overview#nested-routes) and [outlet](https://reactrouter.com/en/main/components/outlet), but I really appreciate your problem solving and making everything work with the current router setup.

- How to enable "hot reload" for this project? Clicking the refresh button in browser to see each css change is a nightmare
I agree. It would be great if we could reduce the hassle of revealing style changes. Maybe this could be an issue that involve some research on configuring webpack and docker? 
 
- Is there a better way to structure or organize the components related to the authentication page?
I wouldn't call it "better" but another approach would be to create a separate Login page and a Signup page, instead of AuthenticationPage + useLocation?

bzzz-coding (approved):
Thanks for creating the login/signup page, @MattPereira ✨ ! I like the idea of creating InputGroup and admire your problem-solving skills 🧠. I also appreciate your comment--let's discuss the next steps and how we can improve our development experience. ❤️

PR #458, opened 2023-10-19
Bug causing deployment to fail with Poetry version > 1.3.0

alexlaw528:
Fixes #457 

### Changes

- Specify version of poetry to download at 1.3.0 as detailed [here](https://github.com/python-poetry/poetry/issues/7611)

PR #461, opened 2023-10-20
Revert "Bug causing deployment to fail with Poetry version > 1.3.0"

MattPereira:
Reverts hackforla/CivicTechJobs#458


alexlaw528 (approved):
Approve revert

PR #615, opened 2024-11-14
Added models and serializers for COP, roles, skills and project

bzzz-coding:
Fixes # 614

### Changes

- Added models for Communities of Practice, Roles, Skills, and Projects
- Added serializer for the models mentioned above
- Checked list of tables in pgdb container to make sure these new tables are shown

### DIFF
--- a/backend/ctj_api/models.py
+++ b/backend/ctj_api/models.py
@@ -18,3 +18,68 @@ class Opportunities(models.Model):
 
     class Meta:
         db_table = "opportunities"
+
+
+class CommunityOfPractice(models.Model):
+
+    class PracticeAreas(models.TextChoices):
+        DATA_SCIENCE = 'data_science', 'Data Science'
+        ENGINEERING = 'engineering', 'Engineering'
+        OPS = 'ops', 'Ops'
+        PROJECT_MANAGEMENT = 'project_management', 'Project/Product Management'
+        UI_UX = 'ui_ux', 'UI/UX'
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    practice_area = models.CharField(max_length=50, choices=PracticeAreas.choices)
+    description = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = "community_of_practice"
+        verbose_name = "Community of Practice"
+        verbose_name_plural = "Communities of Practice"
+
+    def __str__(self):
+        return f"{self.get_practice_area_display()}"
+
+
+class Role(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    title = models.CharField(max_length=50)
+    community_of_practice = models.ForeignKey(CommunityOfPractice, on_delete=models.CASCADE,  related_name="roles")
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = "roles"
+    
+    def __str__(self):
+        return f"{self.title} ({self.community_of_practice.get_practice_area_display()})"
+
+
+class Skill(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    name = models.CharField(max_length=50)
+    roles = models.ManyToManyField('Role', related_name='skills')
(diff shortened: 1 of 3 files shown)


backend/ctj_api/models.py
```diff
@@ -18,3 +18,68 @@ class Opportunities(models.Model):
...
+    class Meta:
+        db_table = "skills"
+
+    def __str__(self):
+        return self.name
+
+
+class Project(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    people_depot_project_id = models.CharField(max_length=255)
```
bzzz-coding, line 76: I added this based on the diagram, but I'm unsure what "people depot project id" is.

LoTerence: This will point to the `project`'s data from the people_depot's database. CTJ is supposed to get it from there.

CTJ has our own database, and people depot has their own database. In the people depot db, the project will have an id as well. `people_depot_project_id` should point to that id as a string. We will use `people_depot_project_id` to easily find the project in people depot db and get its data. It makes it easier to sync our data together.

It's like if we made a textbook database, we would store the ISBN number of the book to make it easier to google that exact book online later. The ISBN comes from an external system, not ours. `people_depot_project_id` is kind of like the project's ISBN number in people depot database.

This app is tricky because some user data needs to be stored in our db, and some needs to be stored in people_depot. Storing the id of the respective user/project from the people_depot database will hopefully help us sync our data with theirs easier.

backend/ctj_api/models.py
```diff
@@ -18,3 +18,68 @@ class Opportunities(models.Model):
...
+class CommunityOfPractice(models.Model):
+
+    class PracticeAreas(models.TextChoices):
+        DATA_SCIENCE = 'data_science', 'Data Science'
+        ENGINEERING = 'engineering', 'Engineering'
+        OPS = 'ops', 'Ops'
+        PROJECT_MANAGEMENT = 'project_management', 'Project/Product Management'
+        UI_UX = 'ui_ux', 'UI/UX'
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    practice_area = models.CharField(max_length=50, choices=PracticeAreas.choices)
```
LoTerence, line 32: I like what you did here with the practice areas 👍

backend/ctj_api/models.py
```diff
@@ -18,3 +18,68 @@ class Opportunities(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = "community_of_practice"
+        verbose_name = "Community of Practice"
+        verbose_name_plural = "Communities of Practice"
+
+    def __str__(self):
+        return f"{self.get_practice_area_display()}"
```
LoTerence, line 43: `self.get_practice_area_display`

What does this function do? I don't see its definition anywhere, unless I missed something 😕

backend/ctj_api/models.py
```diff
@@ -18,3 +18,68 @@ class Opportunities(models.Model):
...
+    people_depot_project_id = models.CharField(max_length=255)
+    meeting_times = models.JSONField(null=True, blank=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = "projects"
+
+    def __str__(self):
+        return self.people_depot_project_id
```
LoTerence, line 85:
```suggestion
        return self.name
```
Oh we should probably keep the name of the project in here 😅 sorry, I will add that to the diagram, so you can use `self.name` instead.

backend/ctj_api/serializers.py
LoTerence, line 1: Nice! This file looks really good 👌

LoTerence (commented):
Code looks great! Just have a couple questions and I think it will be ready to merge :)

Could you also register all the new models in `admin.py` so they can be accessed through the admin panel please?

Thanks for doing this Bitian! It's a huge step toward completing the backend server 🙌

backend/ctj_api/models.py
bzzz-coding: I believe this method is supposed to return the value set in the `choices`. I found the documentation [here](https://docs.djangoproject.com/en/5.1/ref/models/instances/#django.db.models.Model.get_FOO_display).

LoTerence (approved):
Nice, everything  looks great!! Thanks for doing this 🙌

PR #641, opened 2025-02-06
Added Tailwind Button Components

bzzz-coding:
### Changes

- Created button components using Tailwind based on this [Figma design](https://www.figma.com/design/G5bOqhud6azbxyR9El9Ygp/Civic-Tech-Jobs?node-id=12354-1711&t=jvOt1A8v42lzmupy-0)
- Added buttons to `/demo-tailwind` page to showcase enabled, disabled, hover, focus, and active effects
- Added a "Toggle Dark Mode" button to `/demo-tailwind` page to switch between light and dark mode for testing

<details>
  <summary>Light Mode</summary>
  <br>
  
  <br>
</details>

<details>
  <summary>Dark Mode</summary>
  <br>
  
  <br>
</details>

### DIFF
--- a/frontend/src/pages/Demo/DemoTailwind.tsx
+++ b/frontend/src/pages/Demo/DemoTailwind.tsx
@@ -1,14 +1,79 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import Typography from "tw-components/Typography";
+import { Button, IconButton } from "tw-components/Buttons";
 
 const DemoTailwind = () => {
+  // Add a setDarkMode for testing dark mode styles
+  const [darkMode, setDarkMode] = useState(false);
+
+  useEffect(() => {
+    if (darkMode) {
+      document.documentElement.classList.add("dark");
+    } else {
+      document.documentElement.classList.remove("dark");
+    }
+  }, [darkMode]);
+
   return (
     <div className="min-h-screen bg-grey-light p-8">
+      <div className="mx-auto max-w-4xl rounded-lg bg-white p-6 shadow-md dark:bg-[#030D2D] transition-colors duration-300">
+        <Typography.Title1 className="mb-4 text-blue-dark dark:text-white">
+          Buttons
+        </Typography.Title1>
+
+        <Button onClick={() => setDarkMode(!darkMode)}>Toggle Dark Mode</Button>
+
+        {/* Regular Buttons, active is the default state */}
+        <Typography.Title3 className="mt-4 mb-2 text-blue-dark dark:text-white">
+          Active
+        </Typography.Title3>
+        <div className="flex flex-wrap items-center gap-4 mb-4">
+          <Button size="small">Small</Button>
+          <Button size="small-long">Small-Long</Button>
+          {/* Medium is the default*/}
+          <Button>Medium</Button>
+          <Button size="medium-long">Medium-Long</Button>
+          <Button size="large">Large</Button>
+          <Button size="large-long">Large-Long</Button>
+        </div>
+
+        {/* Disabled Buttons */}
+        <Typography.Title3 className="mb-2 text-blue-dark dark:text-white">
+          Disabled
+        </Typography.Title3>
+        <div className="flex flex-wrap items-center gap-4 mb-4">
+          <Button state="disabled">
+            Medium Disabled
+          </Button>
+          <Button size="large" state="disabled">
+            Large Disabled
+          </Button>
+        </div>
+
+        {/* Hover & Active/Focused States */}
+        <Typography.Title3 className="mb-2 text-blue-dark dark:text-white">
+          Hover, Focused and Active
+        </Typography.Title3>
+        <div className="flex flex-wrap items-center gap-4 mb-4">
+          <Button>Hover</Button>
+          <Button>Focus (tab)</Button>
+          <Button>Active (click & hold)</Button>
+        </div>
+
+        {/* Icon Buttons */}
+        <Typography.Title3 className="mb-2 text-blue-dark dark:text-white">
+          Search Icon
+        </Typography.Title3>
+        <div className="flex flex-wrap items-center gap-4">
+          <IconButton />
+          <IconButton state="disabled" />
+        </div>
+      </div>
+
       <div className="mx-auto max-w-4xl rounded-lg bg-white p-6 shadow-md">
         <Typography.Title1 className="mb-4 text-blue-dark">
           Typography Demo (Title 1 - Roboto Bold 48/137% +0)
         </Typography.Title1>
-
         <Typography.Title2 className="mb-2 text-blue-dark">
           Title Styles (Title 2 - Roboto Bold 36/137% +0)
         </Typography.Title2>
--- a/frontend/src/pages/Demo/DemoTailwind.tsx
+++ b/frontend/src/pages/Demo/DemoTailwind.tsx
@@ -42,8 +42,8 @@ const DemoTailwind = () => {
           Disabled
         </Typography.Title3>
         <div className="mb-4 flex flex-wrap items-center gap-4">
-          <Button disabled={true}>Medium Disabled</Button>
-          <Button size="large" disabled={true}>
+          <Button disabled>Medium Disabled</Button>
+          <Button size="large" disabled>
             Large Disabled
           </Button>
         </div>
@@ -64,7 +64,7 @@ const DemoTailwind = () => {
         </Typography.Title3>
         <div className="flex flex-wrap items-center gap-4">
           <IconButton />
-          <IconButton disabled={true} />
+          <IconButton disabled />
         </div>
       </div>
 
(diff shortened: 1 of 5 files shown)


frontend/src/tw-components/Buttons.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const buttonTypography: Record<keyof typeof buttonSizes, React.ElementType> = {
+  small: Typography.Title7,
+  "small-long": Typography.Title7,
+  medium: Typography.Title6,
+  "medium-long": Typography.Title6,
+  large: Typography.Title5,
+  "large-long": Typography.Title5,
+  "icon-only": Typography.Title7, // Not sure about this one
```
LoTerence, line 49: I think the icon-only text size is 24px 🤔 but I'm not sure if I'm reading the figma correctly.

frontend/src/tw-components/Buttons.tsx
```diff
@@ -0,0 +1,100 @@
...
+  size = "medium",
+  state = "enabled",
+  children,
+  onClick,
+}) => {
+  const TextComponent = buttonTypography[size];
+
+  return (
+    <button
+      className={`rounded-[64px] ${buttonSizes[size]} ${buttonColors[state]} transition-all duration-200`}
```
LoTerence, line 62:
```suggestion
      className={`rounded-[64px] ${buttonSizes[size]} ${buttonColors[state]} transition-all duration-200`}
```
Just a suggestion, not a huge deal - Consider using `clsx` to manage tailwind classnames - https://www.npmjs.com/package/clsx

It might help clean up the `Typography.tsx` file as well. But dont worry about it if its overly complicated 😅

frontend/src/tw-components/Buttons.tsx
```diff
@@ -0,0 +1,100 @@
...
+  medium: Typography.Title6,
+  "medium-long": Typography.Title6,
+  large: Typography.Title5,
+  "large-long": Typography.Title5,
+  "icon-only": Typography.Title7, // Not sure about this one
+};
+
+const Button: React.FC<ButtonProps> = ({
+  size = "medium",
+  state = "enabled",
```
LoTerence, line 54:
```suggestion
  disabled = false,
```
Dont pass a prop called `state`, the name is too similar to react's `useState` and component state concept.

Since it can only be either `"enabled"` or `"disabled"`, we can pass a prop called `disabled` whose default value is `false`. What do you think?

frontend/src/tw-components/Buttons.tsx
```diff
@@ -0,0 +1,100 @@
...
+  "large-long":
+    "px-[56px] h-[51px] flex items-center justify-center leading-none",
+  "icon-only": "px-[24px] h-[42px] flex items-center justify-center",
+};
+
+const buttonColors = {
+  enabled:
+    "bg-blue-dark text-white dark:bg-white dark:text-blue-dark hover:bg-blue-dark-hover focus:bg-blue-dark-focused focus:outline-none active:bg-blue-dark-focused dark:hover:bg-grey-light dark:focus:bg-[#D9DBDF] dark:active:bg-[#D9DBDF]",
+  disabled:
+    "bg-grey text-white cursor-not-allowed dark:bg-grey dark:text-grey-light",
```
LoTerence, line 39: There is a way to add classes to disabled buttons in tailwind like this: `disabled:pointer-events-none disabled:opacity-50"`

Maybe we can use something like that instead of an object like `buttonColors`?

frontend/src/pages/Demo/DemoTailwind.tsx
```diff
@@ -1,14 +1,79 @@
...
+    if (darkMode) {
+      document.documentElement.classList.add("dark");
+    } else {
+      document.documentElement.classList.remove("dark");
+    }
+  }, [darkMode]);
+
   return (
     <div className="min-h-screen bg-grey-light p-8">
+      <div className="mx-auto max-w-4xl rounded-lg bg-white p-6 shadow-md dark:bg-[#030D2D] transition-colors duration-300">
```
LoTerence, line 19:
```suggestion
      <div className="mx-auto max-w-4xl rounded-lg bg-white p-6 shadow-md dark:bg-[#030D2D] transition-colors duration-300">
```
If you dont already have it, please get eslint extension on vscode and configure it automatically lint your frontend files.

There is a linter rule for tailwind classnames order that some lines in this file are throwing an error for. You can check by going to the `/frontend/` folder and running `npm run lint`. It should also automatically fix them for you 👍

LoTerence (changes requested):
Overall its really great so far! Thanks for doing this Bitian 😁 I just have a few suggestions above. Please let me know what you think and if you have any questions!

frontend/src/tw-components/Buttons.tsx
bzzz-coding: Thanks for the suggestion! I'll install clsx and try cleaning up the classes a little.

frontend/src/tw-components/Buttons.tsx
bzzz-coding: It's a great idea! I'll make the changes.

frontend/src/tw-components/Buttons.tsx
bzzz-coding: I didn't know about the pointer events. Thanks for the suggestion! I'll change this.

frontend/src/tw-components/Buttons.tsx
bzzz-coding: I thought maybe I could set it to null `"icon-only": null` since it's just an icon and no typography is needed?

frontend/src/pages/Demo/DemoTailwind.tsx
```diff
@@ -41,11 +41,9 @@ const DemoTailwind = () => {
         <Typography.Title3 className="mb-2 text-blue-dark dark:text-white">
           Disabled
         </Typography.Title3>
-        <div className="flex flex-wrap items-center gap-4 mb-4">
-          <Button state="disabled">
-            Medium Disabled
-          </Button>
-          <Button size="large" state="disabled">
+        <div className="mb-4 flex flex-wrap items-center gap-4">
+          <Button disabled={true}>Medium Disabled</Button>
```
LoTerence, line 45:
```suggestion
          <Button disabled>Medium Disabled</Button>
```

We dont need the `={true}`, we can just pass `disabled` by itself.

frontend/src/tw-components/Buttons.tsx
```diff
@@ -1,100 +1,100 @@
...
-          d="m20 20l-4.05-4.05m0 0a7 7 0 1 0-9.9-9.9a7 7 0 0 0 9.9 9.9"
-        />
-      </svg>
-    </button>
-  );
-};
+// Export separate IconButton
+const IconButton: React.FC<Omit<ButtonProps, "size" | "children">> = (
+  props,
+) => <Button {...props} size="icon-only" isIconButton />;
```
LoTerence, line 98:
```suggestion
) => {
  return (
    <Button {...props} size="icon-only" className="rounded-[24px]">
      <svg
        xmlns="http://www.w3.org/2000/svg"
        width="24"
        height="24"
        viewBox="0 0 24 24"
      >
        <path
          fill="none"
          stroke="currentColor"
          strokeLinecap="round"
          strokeLinejoin="round"
          strokeWidth="2"
          d="m20 20l-4.05-4.05m0 0a7 7 0 1 0-9.9-9.9a7 7 0 0 0 9.9 9.9"
        />
      </svg>
    </Button>
  );
};
```

Ideally `<IconButton>` should extend the `<Button>` component, rather than `Button` having a flag to check if its an Icon button or not. 

Otherwise why not just use `<Button isIconButton size="icon-only">` instead of `<IconButton>` in the first place?

With this, we can simplify the Button component by removing the `isIconButton` property.

LoTerence (changes requested):
I like the way you simplified IconButton with this commit so far. There is a little bit more that can be done to make it even cleaner. Good job so far though 😁

frontend/src/tw-components/Buttons.tsx
LoTerence: `cursor-not-allowed` works too and it might be better UI for the skills matrix actually. I just wanted to suggest using the tailwind `disabled:` class instead of making a list of classes for the disabled state. Thats what those examples were about 😅 sorry for the confusion.

Here is a link to what I'm talking about: https://tailwindcss.com/docs/hover-focus-and-other-states#disabled

This way you dont have to make separate variables for storing different classname strings. Your approach is more clear to see whats happening though, so I dont mind keeping that.

frontend/src/tw-components/Buttons.tsx
LoTerence: This will work for now. Lets keep it like that 👍

LoTerence (approved):
Looks great! Thank you for your work on this Bitian!! 

Please go ahead and merge whenever you are ready 😄

PR #646, opened 2025-03-04
Adding Github action to push container to ECR through the Incubator CICD 

RSkuma:
Added ver.1 of the github action, this action pushes a public image from dockerhub and will attempt to verify that the CICD pipeline to incubator is working

### Changes

- adds develop-dev.yml to github actions
- Does not push the application image, but rather a public dockerhub image 
- Once pipeline to confirmed, will update to use our application image and call ECS


LoTerence:
change name to `deploy-stage.yml`

PR #679, opened 2025-04-21
Installed tailwind-merge and implemented cn

### LINKED ISSUE #671: Bug: Resolve Tailwind classnames conflicts
### Story

As developer using tailwind with react, we should have predictable css styling results when using conflicting Tailwind classnames.

### Overview

Example:
- The Checkbox component should have `"cursor-pointer"` tailwind class when enabled, but have `"cursor-not-allowed"` class when `disabled`
- eg. `className={clsx("cursor-pointer", disabled && "cursor-not-allowed")}`
- when `disabled` is true, this gives us a full className string of `"cursor-pointer cursor-not-allowed"`
- The expected result is that `"cursor-not-allowed"` style property would override `"cursor-pointer"`
- The actual result is that no matter what order you put the class names in, `"cursor-pointer"` overrides `"cursor-not-allowed"`. This means it will always be cursor-pointer even when the input is disabled.

This bug would also cause issues when trying to override component styles with `className={clsx("original-tw-classnames", props.className)}`.

See my comment about it here: [Checkbox Component with Tailwind by bzzz-coding · Pull Request #654 · hackforla/CivicTechJobs · GitHub](https://github.com/hackforla/CivicTechJobs/pull/654)

The solution is to implement `tailwind-merge` package to resolve conflicting tailwind classnames like so: [Understanding the Differences Between clsx, classnames, and twMerge in React and Tailwind CSS](https://www.linkedin.com/pulse/understanding-differences-between-clsx-classnames-react-terranova-l5dxf/)

### Action Items

- [x] Create a classname combiner function using `clsx` and `tailwind-merge`
- [x] Implement this new combiner function in every component that passes down `props.className` to override original styles.

Example components: Checkbox component, Button component, CircleCard component

### Resources/Instructions

bzzz-coding:
Fixes #671

### Changes

- Installed tailwind-merge
- Created a cn function in src/lib/utils.ts
- Implement cn function in every tailwind component that passes down props.className


LoTerence (approved):
Looks good! Thanks for doing this Bitian!!

I will go ahead and merge so the rest of us can work on merge conflicts 😁

PR #736, opened 2026-05-02
[2/2] Tailwind + SCSS to CSS Modules

Nickatak:
# PR: Tailwind + SCSS → CSS Modules

**Branch:** `next-rewrite-css-modules` stacked on `next-rewrite` (#735), which is itself stacked on `docs-rewrite` (#734).
**Retarget chain when each parent lands.**

---

## Summary

Removes Tailwind, SCSS, and every utility-class system from the frontend. Components now ship co-located `*.module.css` files; design tokens live as CSS custom properties on `:root`; breakpoints are inlined `@media` queries. PR1 preserved Tailwind + SCSS through the Next.js port so the visual surface stayed put while the framework changed underneath; this is the styling-system swap that PR1 made room for.

This is **PR2 of the 2-PR chain** opened by #735.

## Why this shape

- **Pure CSS, no preprocessor.** Native CSS in 2026 covers everything the existing SCSS used (custom properties, nesting, calc, media queries). The `sass` dep, the `_partials.scss` chain, and the global `@use` plumbing are all gone.
- **Per-feature commits** so each step is small enough to review in isolation. Shared primitives convert first; features pull from those primitives downstream.
- **Co-located modules, no global utilities.** Tailwind-style declarative utility classes (`flex-container`, `col-3`, `paragraph-1`, `mr-4`, etc.) are not preserved as global CSS. Each component's CSS Module owns its layout.

## Commits

| # | Subject |
|---|---------|
| 1  | Drop Tailwind/SCSS toolchain, replace globals with CSS Modules baseline |
| 2  | Convert shared primitives to CSS Modules |
| 3  | Convert shared form/input components to CSS Modules |
| 4  | Convert shared nav + misc components to CSS Modules |
| 5  | Convert shared unused components to CSS Modules |
| 6  | Convert landing feature to CSS Modules |
| 7  | Convert credits feature to CSS Modules |
| 8  | Convert qualifier feature to CSS Modules |
| 9  | Convert session feature to CSS Modules |
| 10 | Convert privacy-policy and not-found features to CSS Modules |

### Toolchain

- **Dropped deps:** `tailwindcss`, `tailwind-merge`, `eslint-plugin-tailwindcss`, `sass`, `autoprefixer`, `postcss`, `postcss-import`.
- **Dropped files:** `tailwind.config.ts`, `postcss.config.mjs`, `src/shared/styles/` (the entire SCSS token + utility partials directory), every `_*.scss` under `src/shared/components/` and `src/features/`.
[...]

PR #735, opened 2026-05-02
[1/2] Port frontend to Next.js 16 (App Router)

Nickatak:
# PR: Port frontend to Next.js 16 (App Router)

**Branch:** `next-rewrite` stacked on `docs-rewrite` (#734).
**Retarget to `develop` once #734 lands.**

---

## Summary

Replaces the legacy `frontend/` (Vite 5 + React 18 + TypeScript 4 + react-router-dom + SCSS + a parallel Tailwind migration) with a Next 16 (App Router) + React 19 + TypeScript 5 + Tailwind 3.4 surface. Behavior preserved; styling carried forward intact so PR2 can sweep Tailwind classes and SCSS into co-located CSS Modules in a clean, isolated diff.

This is **PR1 of a 2-PR chain.** PR2 is Tailwind/SCSS → CSS Modules.

## Why this shape

- **Stacked on docs-rewrite** because the docs in #734 describe the target state we are building. Reviewer can read the spec alongside the code.
- **Two commits (scaffold + port)** rather than dozens of granular ones. Scaffold = the empty Next.js shell + Docker infra; port = all React/TS/test/asset content moving in.
- **Tailwind preserved through the port** (deferred to PR2) so visual surface does not move while the framework changes underneath.

## Commits

| # | Subject |
|---|---------|
| 1 | Scaffold Next.js 16 frontend and split stage Docker |
| 2 | Port legacy frontend to Next.js App Router |

### Stack

- **Next.js 16** (App Router, default Turbopack for dev). Bumped from Next 15 in the docs to match `create-next-app@latest` output.
- **React 19**, **TypeScript 5**.
- **Tailwind 3.4** (carried over). Theme + content paths in `frontend/tailwind.config.ts`. Removed in PR2.
- **SCSS** (carried over). Tokens at `src/shared/styles/`; component-paired files alongside their components.
- **Vitest** + jsdom + @testing-library + vite-plugin-svgr replacing Jest.
- **@floating-ui/react** replacing react-popper (legacy not React-19 compatible).
- **next/font/local** for the bundled Roboto family.

### Architecture

- `src/app/` is route shims only; `src/features/<feature>/` owns real work; `src/shared/` is cross-feature primitives.
- `(with-nav)` and `(auth)` route groups hold the two layout shapes.
- `frontend/ARCHITECTURE_MAP.md` documents layer rules, route-shim policy, controller-API policy, function-style convention, and a route-to-entry table.
[...]

PR #734, opened 2026-04-27
Rewrite docs/ for future-state architecture

Nickatak:
**Stacked on #733.** Merge #733 first; this branch rebases cleanly onto develop afterward.

## Summary

Rewrites the developer documentation for the CTJ rewrite's future-state architecture, flattens `mkdocs/docs/` to `docs/` at repo root, and drops mkdocs entirely.

Architectural direction (per the engineering assessment + meeting alignment):

- **Frontend:** Next.js 15 (App Router) + React 19 + TypeScript 5 + CSS Modules. Replaces the current Vite + React + Tailwind setup.
- **Backend:** Django 5.1 + DRF 3.15 + PostgreSQL 16 (preserved). Stage 1 keeps the data layer local; Stage 2 moves user identity / practice areas / roles / projects to PeopleDepot, with the skill catalog cached locally from PD.
- **Auth:** Stage 1 uses Django's default session auth; Stage 2 adds Cognito ID-token JWT.
- **Topology:** Three containers in one ECS task (Next.js + Django + Postgres-in-container). No RDS.

## Doc structure

`mkdocs/docs/` flattened to `docs/`. mkdocs config / GH Pages workflow / `docker-compose.docs.yml` all removed; GitHub renders the markdown directly.

```
docs/
├── developer/        # 8 active dev docs
├── joining-the-team/ # intro + role-specific guides
├── misc/             # history, research-wiki-template, security-updates
└── resources.md
```

Top-level changes:
- `README.md` rewritten as the canonical project landing (absorbed `mkdocs/docs/index.md` content).
- `CONTRIBUTING.md` rewritten and now includes PR review criteria (absorbed `development-culture.md`).

## Test plan

- [x] Render `README.md`, `CONTRIBUTING.md`, and key docs in `docs/` on GitHub; check formatting and link resolution
- [x] Verify cross-doc links between sibling files (e.g., `backend.md` → `deployment-infra.md`) resolve
- [x] Confirm no broken references to removed files (`mkdocs/`, `index.md`, `development-culture.md`, `the-team.md`, `mkdocs-*.md`)
- [x] Verify `.github/workflows/deploy-stage.yml`'s new `paths-ignore: docs/**` filter is correct intent

## Follow-up cleanup (separate PRs)

- Disable GitHub Pages on the repo (the `gh-pages` branch and `mkdocs-build.yml` workflow are now obsolete)
[...]

PR #737, opened 2026-05-02
Bump dependencies + replace lint stack

Nickatak:
# PR: Bump dependencies + replace lint stack

**Branch:** `next-rewrite-deps-bump` stacked on `next-rewrite-css-modules` (#736), which is stacked on `next-rewrite` (#735), which is stacked on `docs-rewrite` (#734).
**Retarget chain when each parent lands.**

---

## Summary

Modernizes the entire build, test, lint, and CI surface in a single pass. Three commits, each green, read in order:

1. **Bump runtime/framework/library deps.** Python 3.13, Django 6.0, psycopg3, Postgres 18, Node 24, plus the frontend dep tree.
2. **Replace lint stack.** Backend swaps `black + flake8 + isort` for `ruff`, adds `mypy` (gradual) and `bandit`. Frontend keeps ESLint, adds `stylelint` and `knip`. New `lint.yml` CI workflow runs the suite. Pre-commit and docs aligned.
3. **Fix lint workflow CI failures.** Three small fixes for issues that only surfaced on a fresh CI checkout (SVG type declarations, `next typegen`, mypy env vars).

## Why this shape

- **Modernize while the codebase is small.** The rewrite chain (#734, #735, #736) was a deliberately cheap moment to do major bumps; the next forced upgrade in 12-18 months will hit a project with new feature work piled on top.
- **One PR, three commits, "tooling-shaped."** Every change is build / test / lint / CI / dep work; none of it touches business logic. Splitting (e.g. bumps PR + lint-swap PR) was considered but the splits aren't clean: commit 2 deletes black/flake8/isort which commit 1 deliberately *skipped* bumping; knip's dead-code findings entangle with the lint swap; doc alignment for the new tooling can't be decoupled from the tooling itself.
- **Read commit-by-commit.** The unified diff is large (mostly lockfiles, configs, doc updates) but each commit's message tells its own self-contained story.

## Tooling choices

Each new or replaced tool, with the load-bearing alternative we ruled out.

### Backend

- **ruff** (replaces `black + flake8 + isort`). Single Rust binary that does linting, formatting, and import sorting in one pass. Black-compatible formatter plus a rule surface that covers everything the legacy chain did (pyflakes, pycodestyle, isort) plus families the legacy chain needed plugins for (`B` bugbear, `SIM` simplify, `C4` comprehensions, `DJ` Django-specific, `UP` pyupgrade). One config, one CLI, one version pin instead of three. Alternative considered: keep the chain but bump versions. Rejected because the chain itself is the maintenance cost; ruff isn't a marginal upgrade, it's a category change.
- **mypy** (new) in **gradual mode** (`disallow_untyped_defs = false`). The codebase has effectively zero annotations today; strict-from-day-one would force every PR to fight the type checker. Gradual mode runs mypy with django-stubs + drf-stubs so ORM call sites get checked even when surrounding functions are untyped, while the bar to land a typed function stays low. Documented in `CONTRIBUTING.md`. Ratchets stricter as coverage grows. Alternative considered: pyright. Rejected because the Django ecosystem stubs are mypy-first.
- **bandit** (new). Security linter targeting the OWASP-shaped issue classes (hardcoded secrets, SQL injection, weak crypto, unsafe deserialization). Low-friction to add, complements ruff's correctness rules. Configured to skip migrations (auto-generated, lots of false positives) and the test suite (assertions are not security issues in tests).
- **django-stubs + djangorestframework-stubs** (new). The mypy plugins that make Django ORM and DRF call sites type-aware. Without them, mypy treats every `Model.objects.filter(...)` return as `Any` and the type checker is roughly useless on a Django codebase.

### Frontend

- **stylelint** (new) with `stylelint-config-standard`. ESLint can't see CSS files; the post-#736 CSS Modules surface was previously unlinted. Project-specific overrides: camelCase `selector-class-pattern` and `keyframes-name-pattern` (matches `styles.signupForm` JS-import shape), `ignoreProperties` for `composes` (CSS Modules directive that looks like an unknown property to vanilla parsers) and `clip` (canonical `sr-only` recipe).
- **knip** (new) with **fail-on-find** for every category. Detects unused files, unused exports, unused dependencies, and unlisted dependencies. Auto-discovers entry points (App Router pages, configs) and resolves plugin references through their config schemas. Real findings on the first run, all fixed in this PR: 6 unused component files deleted, 3 unused exports removed, 4 unused devDeps removed. Alternatives considered: `depcheck` (only does dependencies), `ts-prune` (only does exports). Knip covers both with one config. Warn-mode was considered and rejected; warnings would accumulate and the whole point of running knip is preventing that.
- **ESLint 9.x** (pinned; defers ESLint 10). `eslint-plugin-jsx-a11y` and `eslint-plugin-react` cap their published peer ranges at `^9`; the maintainers haven't shipped 10-compatible releases. `--legacy-peer-deps` produces a circular-structure JSON error during config loading. Re-evaluation trigger: both plugins shipping `eslint: "^10.0.0"` peers.
- **prettier** (existing, kept). Project default. No project-specific config; defaults are fine. Wired through ESLint via `eslint-plugin-prettier` so formatting failures are lint errors, which means "ready to lint" and "ready to commit" mean the same thing.
- **react-hooks 7.x React-Compiler rules downgraded to `warn`**. The `eslint-plugin-react-hooks` 5 to 7 bump activated three new rules (`set-state-in-effect`, `refs`, `static-components`). The legacy components in this repo violate them in ~28 places (mount-then-render gates, refs read in render, locally-declared subcomponents). Fixing all 29 is a real refactor and out of scope for a "set up the lint suite" PR. Downgraded to `warn` so the signal is visible without blocking merge; intent is to address in a follow-up React-Compiler-prep PR.

### Test runner

- **vitest 3.x** (pinned; defers vitest 4). Vitest 4 requires Vite 8, which requires `@vitejs/plugin-react@6`, which requires `babel-plugin-react-compiler` as a hard peer. We're not using the React Compiler; pulling in a compiler dep just to satisfy a peer range is the wrong trade. Vitest 3.2.4 + plugin-react 5 + Vite 7 is the cleanest tree that supports all our existing test setup. Re-evaluation trigger: `@vitejs/plugin-react` removes the `babel-plugin-react-compiler` peer requirement, or we opt into the compiler.

## Commits

| # | Subject |
|---|---------|
| 1 | Bump runtime, framework, and library dependencies |
| 2 | Replace lint stack: ruff + mypy + bandit (backend) + stylelint + knip (frontend), CI workflow |
| 3 | Fix lint workflow CI failures |

### Commit 1: Dependency bumps

**Backend** (Python 3.12 to 3.13, Django 5.1 to 6.0):

- Django 5.1.2 to 6.0.4
- DRF 3.15.2 to 3.17.1
- daphne 4.1.2 to 4.2.1
- whitenoise 6.8.2 to 6.12.0
- psycopg2-binary 2.9.9 to psycopg[binary] 3.3.4 (driver swap; settings.py needs no change since `django.db.backends.postgresql` auto-detects psycopg3 when installed)
- Python base image 3.12-alpine to 3.13-alpine in both `dev/django.dockerfile` and `stage/django.dockerfile`

**Infra:**

- Postgres 16 to 18 in both compose files (dev volumes need `docker compose down -v` on first pickup)
- Node 22-alpine to 24-alpine in `dev/next.dockerfile` and all three stages of `stage/next.dockerfile`

**Frontend:**

- next 16.2.4 (current), react 19.2.4 to 19.2.5 (patch)
- typescript 5 to 6
- jsdom 25 to 29, prettier 3.4 to 3.8, globals 15 to 17, @types/node 20 to 24
- eslint 9.17 to 9.39 (latest 9.x; ESLint 10 deferred, see Tooling choices)
- eslint-plugin-react-hooks 5.1 to 7.1
- @vitejs/plugin-react 4 to 5
- vitest 2 to 3 (vitest 4 deferred, see Tooling choices)
- `vitest.config.ts` renamed to `vitest.config.mts` since `@vitejs/plugin-react@5` is ESM-only

Lint tooling (black, flake8, isort) intentionally left at current versions; those are replaced wholesale in commit 2.

### Commit 2: Lint stack replacement

**Backend** (swap `black + flake8 + isort` for `ruff`):

- `pyproject.toml`: replaces three dev deps with ruff under `[tool.ruff]` / `[tool.ruff.lint]` (selecting `E/W/F/I/B/UP/DJ/SIM/C4`), `[tool.mypy]` (gradual mode with django-stubs + drf-stubs plugins), `[tool.django-stubs]` (settings module pointer), `[tool.bandit]` (excludes migrations + tests).
- `backend/.flake8` deleted; ruff config supersedes it.
- ruff format pass produced trivial reshape on `manage.py` and `settings.py` (line-length-driven splits).
- Three pre-existing ruff violations addressed:
  - **E501** in `settings.py` (line over 88 chars). Split.
  - **DJ008** on `Opportunity` model (no `__str__`). Added `f"{role.title} @ {project.name}"`.
  - **DJ001** on `min_experience_required` (CharField with `null=True`). `noqa`'d with TODO comment. The fix drops `null=True` which generates a schema migration; deferred to a focused follow-up PR rather than expanding this PR's scope.

**Frontend** (extend the lint surface beyond ESLint):

- ESLint stays at 9.x; `eslint-config-next` 16's flat configs are loaded directly (no more `FlatCompat` shim).
- The pre-rewrite `eslint.config.mjs` referenced `eslint-plugin-tailwindcss` even though Tailwind was deleted in #736. Removed.
- The legacy `next lint` script no longer exists in Next 16; lint script now runs `eslint .` directly.
- `eslint-plugin-react-hooks` 5 to 7 brings the React Compiler rule set; three new rules downgraded to `warn` (see Tooling choices).
- `stylelint` added with `stylelint-config-standard` plus camelCase patterns and ignore lists (see Tooling choices).
- `knip` added with fail-on-find for every category. Real findings in this PR:
  - 6 unused component files deleted (`AccordionFaq`, `ClickCarousel`, `ScrollCarousel`, `ChevronScroll`, `ChipsSelection`, `Chip`)
  - 3 unused exports removed (`SearchButton`, `sampleCopData`, `combineClasses`)
  - 1 unused type re-export removed (`AssetDatum` from `creditsIconData.ts`)
  - 4 unused devDependencies removed (`@eslint/js`, `globals`, `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `typescript-eslint`; the latter three were transitive via `eslint-config-next`)
- `@svgr/webpack` stays in `package.json` since `next.config.ts` references it as a string loader name; knip can't see that through configs, so it's the sole entry in `ignoreDependencies`.

**Pre-commit:**

- `.pre-commit-config.yaml` rewritten. The previous file was stale: pinned to Python 3.9, referenced paths that no longer exist (`app/setup.cfg`, `config.settings`), ran old versions of the tools we're now deleting. New config: `ruff-pre-commit` (check + format passes), `mirrors-prettier`, local hooks for eslint and stylelint that delegate to `npx` so the project's pinned versions are used. Python pinned to 3.13, Node to 24.

**CI:**

- `.github/workflows/lint.yml` added. Two parallel jobs (backend / frontend) each running their respective full lint suites. Authoritative gate; PRs cannot merge while red.
- `.github/workflows/jest-react-test.yml` renamed to `vitest-test.yml`; Node 18 to 24.
- `.github/workflows/linter.yml` (Super-Linter) deleted; superseded.

**Cleanup:**

- `dev/linter.dockerfile` + `dev/linter.env.example` deleted. The dev linter container was the legacy enforcement path for the old toolchain; CI workflow + pre-commit replace it.

**Docs (single source of truth alignment):**

- `docs/developer/eslint-guide.md` renamed to `frontend-lint-guide.md` and rewritten to cover the four-layer frontend lint stack (ESLint + Stylelint + Knip + Prettier) plus tsc.
- `docs/developer/backend.md`: lint section updated; directory tree no longer lists `.flake8`.
- `docs/developer/quickstart-guide.md`: backend + frontend lint command sections updated.
- `docs/developer/devops.md`: linting section rewritten with a tool/surface table.
- `docs/developer/installation.md`: required Node 22 to 24, Python 3.12 to 3.13, pre-commit Python hooks list updated to ruff. Editor-extension recommendations expanded (ESLint, Stylelint, Prettier).
- `docs/developer/design-system.md`, `docs/resources.md`, `README.md`: link updates from eslint-guide to frontend-lint-guide.
- `CONTRIBUTING.md`: typing-policy section added describing the gradual mypy posture. New code annotates signatures, existing code is annotation-welcome but not required, don't annotate just to silence the type checker.

### Commit 3: CI workflow fixes

Three issues surfaced when commit 2's `Lint` workflow ran on a fresh CI checkout (none reproducible locally because of cached `.next/` and Docker env vars):

- `tsc --noEmit` couldn't resolve `@/shared/icons/*.svg` and `@/shared/images/*.svg` imports. SVGR transforms `.svg` into React components at runtime but doesn't emit `.d.ts` files. Added `frontend/src/types/svg.d.ts` with the ambient module declaration.
- `tsc` also failed on `next-env.d.ts`'s static import of `./.next/types/routes.d.ts`, which doesn't exist on a fresh checkout. The `lint:types` script now runs `next typegen && tsc --noEmit`.
- `mypy` crashed inside django-stubs because `backend.settings` reads required env vars via python-decouple at import time. Added dummy env vars to the mypy step in `lint.yml`.

## Test plan

- [x] `ruff check`, `ruff format --check`, `mypy`, `bandit` all clean
- [x] `eslint`, `stylelint`, `knip`, `tsc --noEmit`, `prettier --check` all clean
- [x] Backend `python manage.py test`: 9 tests pass on Django 6 / Python 3.13 / psycopg3 / Postgres 18
- [x] Frontend `vitest run`: 11 tests pass, 3 skipped (matches pre-PR baseline)
- [x] Frontend `next build` succeeds
- [x] `docker compose up --watch` brings up Django 6 + Daphne 4.2 cleanly on Postgres 18
- [x] CI: `Lint` and `Vitest` workflows both green

## Follow-ups (not blocking this PR)

- **Schema migration**: drop `null=True` on `Opportunity.min_experience_required` (DJ001 violation `noqa`'d here). Generates a migration; deserves its own focused PR.
- **React-Compiler prep**: address the 28 react-hooks rule warnings. Real refactor, restructuring mount-then-render gates, redesigning Calendar's drag tracking, hoisting locally-declared subcomponents.
- **GitHub Actions deprecation**: `actions/checkout@v4` and `actions/setup-python@v5` show a non-blocking deprecation warning about Node 20 (forced to Node 24 in June 2026). Bump when convenient.
- **Re-evaluate ESLint 10** once `eslint-plugin-jsx-a11y` and `eslint-plugin-react` ship 10-compatible releases.
- **Re-evaluate vitest 4** once `@vitejs/plugin-react` no longer requires `babel-plugin-react-compiler` as a hard peer (or the project opts into the compiler).


RSkuma (approved):
Approved after discussion

PR #738, opened 2026-05-02
docs: Establish branch and commit message conventions

Nickatak:
**Branch:** `docs/git-conventions` stacked on `next-rewrite-deps-bump` (#737), which is itself stacked on the rewrite chain (#734-#736).
**Retarget chain when each parent lands.**

---

## Summary

Two small doc updates that make branch naming and commit message prefixes a requirement rather than a suggestion. Forward-only; existing branches and history are grandfathered.

### Branch naming (`docs/developer/git-branch-structure.md`)

The existing "Naming" section said prefixes are optional. Rewritten to require a `<type>/<short-description>` shape with five types:

| Prefix | When |
|--------|------|
| `feat/` | New feature, new endpoint, new component |
| `fix/` | Bug fix |
| `chore/` | Dependency bumps, tooling, CI, build, formatting |
| `docs/` | Documentation only |
| `refactor/` | Restructuring without behavior change |

`ci/`, `build/`, `style/`, `test/` are deliberately subsumed under `chore/` so contributors don't second-guess between near-synonyms.

Examples added inline: `feat/auth-stage1`, `fix/dropdown-keyboard-nav`, `chore/bump-django`, `docs/api-reference`, `refactor/extract-skill-matrix`.

### Commit messages (`CONTRIBUTING.md`)

New "Commit messages" section, placed between "Submitting a pull request" and "Code review". Mirrors the same five types as a `<type>: <description>` title shape, optional scope (`feat(auth): ...`), imperative mood, sentence case, no trailing period.

Body conventions kept as-is: prose explaining *why*, not *what*; the diff already shows the what.

## Why now

The rewrite chain (#734-#737) was infrastructure-shaped: most of it didn't fit cleanly under any single conventional prefix, so adopting the convention mid-chain would have produced misleading labels. Establishing it after the chain settles, before the first feature work, lets new commits use the convention from day one.

## Forward-only

Convention applies to new branches and new commits. Existing chain branches (`docs-rewrite`, `next-rewrite`, `next-rewrite-css-modules`, `next-rewrite-deps-bump`) keep their existing names; PR titles in those PRs aren't being rewritten. Renaming live PRs and rewriting history is more churn than the consistency is worth.

## Enforcement

Doc + PR review for now. A branch-name CI check or a commit-msg hook would be overkill for a two-person team; the convention is small enough to remember.

## Test plan

- [x] Both files render in the GitHub UI without broken markdown
- [x] Cross-link from CONTRIBUTING.md to the Naming section in git-branch-structure.md resolves

## Follow-ups

- None blocking. If the convention turns out to be too loose or too strict in practice, revisit and update the doc; the cost of changing a doc-only convention is low.


RSkuma (approved):
Free lunch, approved

PR #739, opened 2026-05-02
docs: Add docstring conventions and apply across backend + frontend

Nickatak:
# docs: Add docstring conventions and apply across backend + frontend

**Branch:** `docs/add-docstrings` stacked on `docs/git-conventions` (#738), which is stacked on `next-rewrite-deps-bump` (#737), which is stacked on the rewrite chain (#734-#736).
**Retarget chain when each parent lands.**

This is the first **pattern-establishment** PR after the modernization chain. The post-modernization tip is the cheapest moment to lock in wide-domain conventions before new feature work starts; this PR locks in docstrings so subsequent pattern PRs (view shape, serializer shape, etc.) can reference an established docstring style as they land.

---

## Summary

Establishes per-file-kind docstring conventions and applies them across the entire `ctj_api` backend (10 source files plus tests + project scaffold) and the entire frontend `src/` tree plus tests (54 files). No behavior changes; this PR is pure documentation.

77 files changed, 2,085 insertions, 171 deletions. Ten commits, one per coherent batch.

| # | Subject |
|---|---------|
| 1 | docs: Add backend docstring style guide and typing convention |
| 2 | docs: Add docstrings to ctj_api models and views |
| 3 | docs: Add docstrings to ctj_api serializers and permissions |
| 4 | docs: Add docstrings to backend scaffold and config files |
| 5 | docs: Add frontend docstring style guide |
| 6 | docs: Add docstrings to frontend foundational files |
| 7 | docs: Add docstrings to frontend app routes |
| 8 | docs: Add docstrings to frontend shared components |
| 9 | docs: Add docstrings to frontend feature components |
| 10 | docs: Add docstrings to frontend tests |

## Why this shape

- **Conventions land first** (commits 1 and 5). Each subsequent commit references the established template in `docs/developer/`, so reviewers can see the rule and the application together.
- **Backend before frontend.** Backend docstrings have substantive policy to encode (lifecycle, visibility, permissions); the labeled-section template developed there informed the lighter frontend convention. Locking the heavier shape first kept the constraints honest.
- **Per-file-kind grouping.** Each commit covers one logical batch (models + views, serializers + permissions, app routes, shared components, etc.) so reviewers can read commit-by-commit and see the same template applied consistently within each.
- **Pure-docs PR by intent.** Bugs surfaced during the sweep are flagged in docstrings (where they're durable and grep-friendly) but not fixed. See "Bugs flagged" below.

### Backend convention - heavy labeled sections for substantive files

[...]

PR #740, opened 2026-05-02
refactor: Convert single-purpose views to function-based views

Nickatak:
# refactor: Convert single-purpose views to function-based views

**Branch:** `refactor/views-explicit` stacked on `docs/add-docstrings` (#739), which is stacked on the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This is the second pattern-establishment PR after the modernization chain. With docstring conventions locked in by #739, this PR locks in the **view shape** convention: full-CRUD resources stay as `ModelViewSet`s on `DefaultRouter`, and everything narrower (single method, list-only, retrieve-only, list+retrieve) becomes a function-based view (FBV) with explicit `path()` routing.

---

## Summary

Converts `UserDetail(RetrieveAPIView)` and the four `ReadOnlyModelViewSet`s (`CommunityOfPractice`, `Role`, `Skill`, `Project`) to function-based views. Routes in `ctj_api/urls.py` move from `DefaultRouter` auto-registration to explicit `path()` entries.

`OpportunityViewSet` stays as `ModelViewSet` (full CRUD on the recruitment catalog). `healthcheck` and `api_not_found` remain plain Django views.

3 files changed, 222 insertions, 78 deletions. Two commits.

| # | Subject |
|---|---------|
| 1 | refactor: Convert single-purpose views to function-based views |
| 2 | docs: Document view shape convention in backend.md |

## Why this shape

The class-based shapes hide behavior. `ReadOnlyModelViewSet` resolves to "list + retrieve" only if you know DRF; nothing in the class body says so. `RetrieveAPIView` resolves to "GET single record" only if you know DRF. Routes are auto-generated by `DefaultRouter` from class names + `basename`, so nothing in `views.py` tells you what URL the view answers on. Reading `views.py` cold, you can't answer "what methods does this view expose?" or "what URL serves it?" without cross-referencing DRF source and `urls.py`.

After this PR:

- `views.py` is two shapes instead of three: `ModelViewSet` for full CRUD, FBV for everything narrower. Plain Django views for non-resource endpoints.
- Every FBV declares its method (`@api_view(["GET"])`) and permissions (`@permission_classes([...])`) inline.
- Every URL in `urls.py` is grep-able as a literal `path()` entry. `grep "/api/skills"` lands you on a real line.

The convention is documented in [docs/developer/backend.md](https://github.com/hackforla/CivicTechJobs/blob/main/docs/developer/backend.md) (new `View shape` section) and encoded in the module docstring at the top of `ctj_api/views.py`:

| Endpoint shape | View shape |
|----------------|------------|
| Full CRUD (>=4 actions of list/create/retrieve/update/destroy) | `ModelViewSet` registered on `DefaultRouter` |
| Anything narrower | FBV with `@api_view([...])` and explicit `permission_classes` |
| Non-resource (health, fallbacks) | Plain Django `def view(request)` |

### `has_object_permission` in FBV land

`UserDetailPermission.has_object_permission` was previously auto-triggered by `RetrieveAPIView`'s `get_object()` machinery. FBVs don't have an APIView class to dispatch through, so the check has to be invoked explicitly:

```python
@api_view(["GET"])
@permission_classes([permissions.IsAuthenticated, UserDetailPermission])
def user_detail(request, pk):
    user = get_object_or_404(CustomUser, pk=pk)
    request.parser_context["view"].check_object_permissions(request, user)
    serializer = CustomUserSerializer(user)
    return Response(serializer.data)
```

The `request.parser_context["view"]` is the wrapper APIView that `@api_view` synthesizes; `check_object_permissions` is the same method that the generic views call internally. The pattern is documented in DRF, but worth flagging since it's the only inline divergence from the read-only FBV shape.

A WHY-comment in the view body marks this so readers don't reach for "why isn't this auto-triggered like in the generic?"

### URL ordering

`urls.py` routing order:

1. `healthcheck` (single fixed path).
2. `users/<uuid:pk>/` (FBV).
3. List + detail FBV pairs for the four read-only catalogs.
4. `re_path(r"^", include(router.urls))` for `OpportunityViewSet`.
5. Catch-all `re_path(r"^.*$", views.api_not_found)`.

The router include sits before the catch-all so `/api/opportunities/...` resolves to the viewset before the catch-all gets a shot. The explicit FBV paths sit before the router include for the same reason — none of them collide with router-managed paths, but the explicit-first order keeps the routing rules predictable.

## Test plan

This is a shape-only change; no behavior changes are intended.

  - Test count is 9 tests in `test_api_endpoints.py` (one is commented out pending BUG-001 fix).
- [x] URL surface unchanged: every path the viewsets used to expose is still served by the FBV pair, including the camelCase `/api/communityOfPractice/` path.

## Follow-ups (not blocking)

The bug-fix PRs flagged in #739 are still next in the queue:
- Architectural cleanup (SPA-catchall remnant, `frontend_dist` references, whitenoise workaround).
- Permission-pattern PR (`OpportunityPermission` PATCH gap).
- Serializer-pattern PR (broken `opportunities` field, dead `SkillMatrixSerializer`).
- A11y / UX polish PR (Dialog, hamburger menu, sitemap, fade animation, hard-coded localhost).

- **Serializer shape** — current serializers mix `ModelSerializer` patterns with one explicit `Serializer`; lock in a convention.
- **Error envelope** — `api_not_found` defines a JSON error shape that nothing else follows; either lift it to a project-wide convention or document why the catch-all is special.
- **Test shape** — the 21 tests are organized as one big `APIBasicTests` class; lock in per-resource test class organization.
- **URL routing convention** — the camelCase `/api/communityOfPractice/` is inconsistent with the snake_case-ish `/api/community_of_practice/` Pythonic shape; needs a decision.
- **Permission class shape** — the two existing permission classes diverge in style (`OpportunityPermission` branches both methods; `UserDetailPermission` only `has_object_permission`); lock in a convention.

Each is independent of the others, so they can land in any order against `develop`.


RSkuma (approved):
Approved

PR #743, opened 2026-05-03
refactor: Apply kebab-case + trailing-slash URL routing convention

Nickatak:
# refactor: Apply kebab-case + trailing-slash URL routing convention

**Branch:** `refactor/url-routing` stacked on `refactor/test-shape` (Pattern PR #4, draft pending), which is stacked on `refactor/serializers-read-write` (Pattern PR #3, draft pending), which is stacked on `refactor/views-explicit` (#740), `docs/add-docstrings` (#739), and the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This is the fifth pattern-establishment PR. With docstrings, view shape, serializer shape, and test shape locked in, this PR locks in the **URL routing convention**: resource URLs are kebab-case, plural, and trailing-slashed.

---

## Summary

- Renames `/api/communityOfPractice/` (camelCase, only outlier) to `/api/communities-of-practice/`.
- Adds trailing slash to `/api/healthcheck` → `/api/healthcheck/` (the only slash-less route).
- Updates inline docstring URL references across `views.py`, `serializers.py`, and `models.py`.
- Updates the two test files that exercise the renamed routes.
- Documents the convention in [`docs/developer/backend.md`](https://github.com/hackforla/CivicTechJobs/blob/main/docs/developer/backend.md) (new `URL routing` section) and refreshes the `API endpoints` table to reflect actual routes.

7 files changed, 35 insertions, 17 deletions. Two commits.

| # | Subject |
|---|---------|
| 1 | refactor: Apply kebab-case + trailing-slash URL convention |
| 2 | docs: Document URL routing convention |

## Why this shape

`urls.py` at the post-test-shape tip mixed three URL styles in eight routes:

- `/api/skills/`, `/api/roles/`, `/api/projects/`, `/api/opportunities/`, `/api/users/<uuid:pk>/` — single-word or snake-equivalent, plural where applicable, trailing slash. (Most routes.)
- `/api/communityOfPractice/` — **camelCase, singular, trailing slash**.
- `/api/healthcheck` — single word, **no trailing slash**.

There was no documented convention; route shapes were inherited from prior contributors. The mismatch is small today (one camelCase URL, one missing-slash URL), but the codebase is about to add many endpoints (auth rebuild, matching, qualifier integration). Locking in a rule now is cheaper than enforcing one after a wave of new routes lands.

## The convention

| Aspect | Rule | Example |
|--------|------|---------|
| Word casing | kebab-case (`-` between words) | `communities-of-practice`, `forgot-password` |
| Pluralization | Plural for collection endpoints; singular for non-resource | `/api/skills/`, `/api/healthcheck/` |
| Trailing slash | Always present | `/api/skills/`, not `/api/skills` |
| Path parameters | UUID via `<uuid:pk>`; named when multi-segment (`<uuid:project_id>`) | `/api/users/<uuid:pk>/` |

Choices in brief:
- **kebab-case over snake_case**: URLs are protocol-level path segments, not Python identifiers. The HTTP/REST tradition is kebab; underscores can disappear under hyperlink underline. BnC also uses kebab-case across its routes.
- **kebab-case over camelCase**: camelCase in URLs is uncommon and was the current outlier, not a chosen style.
- **Always trailing slash**: Django's `APPEND_SLASH=True` (the default) issues a 301 from `/api/skills` to `/api/skills/`; matching the redirect target eliminates the round-trip.
- **Plural for collections**: matches DRF router defaults, BnC, and the rest of the existing codebase.

## Frontend impact: zero

Audit before the rename: `grep -rn "/api/" frontend/src/` returns only one match — a comment in `copData.ts` mentioning that the data could "be replaced with a fetch from `/api/skills/`". No actual fetch calls exist. The qualifier UI exists but doesn't yet submit, and the auth fetch wrapper lives in the discarded `archive/feat-auth-stage1` branch.

The renamed `/api/communities-of-practice/` and `/api/healthcheck/` URLs have no callers outside this repo. The rename is atomic and zero-blast-radius.

Following the same pure-pattern rule as the prior PRs:

- No new routes added; no routes removed (only rename).
- No view-layer or serializer-layer behavior changes.
- No model field renames (`CommunityOfPractice` the model class stays; only the URL changes).
- API versioning is not introduced (`/api/v1/...` pattern explicitly out of scope; left as a future decision).
- The pre-existing inaccuracies in `docs/developer/backend.md` (references to `/api/users/me/` and `UserProfilePermission` which don't exist in the current code) are partially cleaned up in the API endpoints table; broader doc accuracy is a separate concern.

## Test plan

- [x] Frontend impact audit: no `/api/` fetch calls exist (only one comment reference, unrelated to renamed URLs)

## Follow-ups (not blocking)

- **Permission class shape** — `OpportunityPermission` branches both methods; `UserDetailPermission` only `has_object_permission`; lock in a convention.
- **Error envelope** — `api_not_found` defines a JSON error shape that nothing else follows; partly inventing the convention.

The bug-fix and architectural-cleanup PRs from #739 are still queued.

---


RSkuma (approved):
Approved, standardization and consistency is good

PR #744, opened 2026-05-03
docs: Document permission class shape

Nickatak:
# docs: Document permission class shape

**Branch:** `refactor/permission-shape` stacked on `refactor/url-routing` (Pattern PR #5, draft pending), which is stacked on `refactor/test-shape`, `refactor/serializers-read-write`, `refactor/views-explicit` (#740), `docs/add-docstrings` (#739), and the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This is the sixth pattern-establishment PR. It locks in the **permission class shape**: when to override which DRF method, how to structure method bodies, and how to compose `permission_classes` in views.

This PR is **docs-only**. The two existing permission classes (`OpportunityPermission`, `UserDetailPermission`) already conform to the documented convention; this PR writes down what they're already doing so the next class to land has a written rule to follow.

---

## Summary

- Adds a `Permission class shape` subsection inside the existing `Permissions` section in [`docs/developer/backend.md`](https://github.com/hackforla/CivicTechJobs/blob/main/docs/developer/backend.md).
- Documents three rules: override only the layers (`has_permission` / `has_object_permission`) your policy needs; method-body conventions (early-return branches, `getattr` for AnonymousUser-safety, SAFE_METHODS exemption only when the resource is publicly readable); `permission_classes` ordering (built-ins first, custom after; tuple in class attr, list in decorator).
- Corrects two stale references: `UserProfilePermission` (doesn't exist) → `UserDetailPermission`; `is_project_manager` (wrong casing) → `isProjectManager` (matches the model field).
- Updates the user-detail route reference to the current `/api/users/<uuid:pk>/` shape.

1 file changed, 25 insertions, 2 deletions. One commit.

| # | Subject |
|---|---------|
| 1 | docs: Document permission class shape |

## Why this is docs-only

`OpportunityPermission` overrides both `has_permission` and `has_object_permission` because its policy spans both layers (HTTP-method gates: only PMs can POST; per-row gates: only the creator can PUT). `UserDetailPermission` overrides only `has_object_permission` because its entire policy is identity-matching (own-profile-only) which is fundamentally per-row.

Reading the two classes side-by-side, the convention is already there — but unwritten. The eventual contributor reads `permissions.py` cold, sees one class with two methods and one class with one method, and asks "why?" — better to have a written answer in `backend.md` than to encode the answer in tribal knowledge.

The PATCH gap in `OpportunityPermission` (BUG-003 from #739: `PATCH` requests are always 403'd because neither `has_permission` nor `has_object_permission` has a PATCH branch) is unchanged here. That's the permission-bug-fix PR's job, not the permission-shape PR's.

## The convention

**Override only the layers your policy needs:**

- `has_permission` for rules that depend only on the request (HTTP method, user properties readable without a row lookup, headers).
- `has_object_permission` for rules that depend on the row being checked (creator-only, organization-membership, status-based).
- Both when rules span both layers.
- Neither — use a built-in DRF class (`IsAuthenticated`, `IsAuthenticatedOrReadOnly`, `AllowAny`) — when the policy reduces to one of those. Custom classes are for *resource-specific* policy.

**Class naming:**

- `<Resource>Permission` for resource-scoped policy (`OpportunityPermission`, `UserDetailPermission`).
- `Is<Capability>Permission` for cross-resource capabilities (none in the codebase yet, but reserved for the auth rebuild's `IsProjectManager`-style classes).

**Method-body conventions:**

- Early-return branches rather than nested `if`s. Each method ends with explicit `return False` for fall-through.
- `getattr(request.user, "attr", default)` for fields anonymous users don't have. Direct `request.user.isProjectManager` raises `AttributeError` on `AnonymousUser`.
- SAFE_METHODS exemption (`if request.method in permissions.SAFE_METHODS: return True`) only when the resource is publicly readable. For self/owner-only resources, do not exempt SAFE methods — the identity check handles all methods.
- Inline comments at branch boundaries when the branch's intent isn't obvious from the code; skip comments for one-line obvious branches.

**`permission_classes` composition (in views):**

- Tuple for class-level attributes: `permission_classes = (IsAuthenticatedOrReadOnly, OpportunityPermission)`.
- List for the `@permission_classes([...])` decorator on FBVs (Python convention for decorator arguments).
- Order from least-to-most-specific: DRF built-ins first (auth), resource-specific custom classes after. Reading top-to-bottom reads the gate progression.

- `permissions.py` is untouched. The existing two classes already conform to the convention.
- The PATCH gap on `OpportunityPermission` (BUG-003) stays. The bug-fix PR (queued separately) drops it.
- No test changes; no behavior changes.

## Test plan

- [x] `ruff check ctj_api/ backend/` clean (no code changes anyway)
- [x] `mypy ctj_api/ backend/` clean
- [x] All 9 existing backend tests pass (same as before — no behavior changes)

## Follow-ups (not blocking)

The remaining pattern-PR axis:

- **Error envelope** — `api_not_found` defines a JSON error shape that nothing else follows; the next PR locks in a project-wide convention.

The bug-fix and architectural-cleanup PRs from #739 are still queued.

---


RSkuma (approved):
Approved

PR #745, opened 2026-05-03
refactor: Adopt single error-envelope shape via custom exception handler

Nickatak:
# refactor: Adopt single error-envelope shape via custom exception handler

**Branch:** `refactor/error-envelope` stacked on `refactor/permission-shape`, which is stacked on `refactor/url-routing`, `refactor/test-shape`, `refactor/serializers-read-write`, `refactor/views-explicit` (#740), `docs/add-docstrings` (#739), and the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This is the seventh and final pattern-establishment PR. It locks in the **error envelope shape**: a single JSON object structure used by every error response, applied automatically to DRF-raised exceptions via a custom handler and inline at the one manual error path (`api_not_found`).

---

## Summary

- New `ctj_api/exceptions.py` with `civic_exception_handler` that wraps DRF's default exception handler. Registered as `REST_FRAMEWORK["EXCEPTION_HANDLER"]` in `backend/settings.py`.
- `api_not_found` (the catch-all 404 for unknown `/api/*` paths) rewritten to use the envelope shape inline — it's a Django URL fallback, not a DRF error path, so it doesn't go through the handler.
- Two existing 403 tests extended to assert envelope shape; new `test_errors.py` adds one test for `api_not_found`'s envelope path.
- Documents the convention in [`docs/developer/backend.md`](https://github.com/hackforla/civictechjobs/blob/main/docs/developer/backend.md) (new `Error envelope` section).

6 files changed, 185 insertions, 5 deletions. Two commits.

| # | Subject |
|---|---------|
| 1 | refactor: Adopt single error-envelope shape via custom exception handler |
| 2 | docs: Document error envelope shape |

## Why this shape

`api_not_found` had defined a one-off shape (`{"error": "...", "status_code": ..., "message": "..."}`) for the catch-all 404 path. Every other error response was DRF's default — `{"detail": "..."}` for permission/auth/not-found, `{"<field>": [...]}` for per-field validation errors, `{"non_field_errors": [...]}` for top-level validation errors. A frontend caller had to handle three different error shapes, and the catch-all's `status_code` field duplicated information already in the HTTP response status.

This is the final pattern PR. With every other axis (docstrings, view shape, serializer shape, test shape, URL routing, permission shape) locked in, the error envelope was the last remaining axis where the codebase exposed multiple shapes for the same kind of thing. Locking in one shape now means the auth rebuild, matching endpoint, and qualifier integration all start from the same convention.

## The envelope

```json
{
  "error": {
    "code": "<machine_readable_snake_case>",
    "message": "<human-readable string>",
    "fields": { "<field>": ["<msg>"] }
  }
}
```

- `error` is always an object, never a string.
- `error.code` is a snake_case machine-readable identifier. Stable across translations; clients switch on it.
- `error.message` is a human-readable string suitable for displaying to users when no field-level surface is appropriate. Punctuation included.
- `error.fields` is present **only** for validation errors. Keys are field names, values are arrays of messages. Top-level (non-field) validation errors land under `non_field_errors`, mirroring DRF's existing convention.
- HTTP status code is **not** duplicated in the body — it lives only in the response status. (The previous catch-all's `status_code: 404` field is removed; clients have the status from the HTTP response line.)

DRF exception → code mapping in `civic_exception_handler`:

| Exception | `error.code` |
|-----------|--------------|
| `ValidationError` | `validation_error` (with `error.fields`) |
| `AuthenticationFailed` | `authentication_failed` |
| `NotAuthenticated` | `not_authenticated` |
| `PermissionDenied` | `permission_denied` |
| `NotFound` | `not_found` |
| `MethodNotAllowed` | `method_not_allowed` |
| `NotAcceptable` | `not_acceptable` |
| `UnsupportedMediaType` | `unsupported_media_type` |
| `ParseError` | `parse_error` |
| `Throttled` | `throttled` |
| (other DRF `APIException`) | `error` |

## Why a handler (and not BnC-style inline)

BnC constructs envelopes inline at every error path; CTJ uses a custom handler. The two approaches reflect different codebase shapes:

- BnC has many manually-constructed error paths (domain validation, signing flow guards, customer-email-required gates) where the error happens at a specific business-logic checkpoint, not at the framework boundary. Inline construction reads naturally there.
- CTJ's errors mostly come from DRF's machinery (auth, permission, validation, not-found, method-not-allowed). Without a handler, every viewset/FBV that *could* fail would need its own try/except wrapping. The cost is paid at every endpoint instead of once at the handler.

Both approaches produce the same shape for the consumer; they differ in where the shape lives. CTJ's handler is the right shape for CTJ's path mix today.

The one place CTJ does construct an error inline is `api_not_found`, because it's reached as a Django URL fallback (the catch-all `re_path(r"^.*$", api_not_found)` in `ctj_api/urls.py`) rather than through DRF's exception machinery. The handler can't reach it; the inline construction uses the same envelope shape so the contract is uniform.

## Test plan

- [x] All 10 backend tests pass (9 from before + 1 new envelope test for `api_not_found`)
- [x] Two existing 403 tests now additionally assert envelope shape (`permission_denied` and `not_authenticated` codes resolve correctly through the handler)

- No new endpoints; no behavior changes beyond response body shape.
- No bug fixes from the queue.
- The catch-all behavior (404 for unknown `/api/*` paths) is unchanged — only its response body shape changes.
- DRF's per-field validation detail format (`{"field_name": [...]}`) is preserved as the value of `error.fields` rather than re-keyed; clients keep the same field-level structure.

## Follow-ups (not blocking)

This is the last pattern-axis PR. After this lands, the queue moves to bug-fix and architectural-cleanup PRs:

- **Architectural cleanup** clusters from #739 (drop SPA-catchall remnant + `frontend_dist` references + whitenoise workaround).
- **Permission-bug-fix PR** resolves the `OpportunityPermission` PATCH gap.
- **Serializer-bug-fix PR** drops the broken `opportunities` field and the dead `SkillMatrixSerializer`.
- **A11y / UX polish PR** covers the frontend bugs flagged in #739.
- **Eventual auth rebuild** uses `scratch/planning/planned_auth.md` as the spec, applying every pattern locked in by this stack.

---


RSkuma (approved):
Approved, discussed, free lunch

PR #746, opened 2026-05-03
docs: Add public ADR series in docs/decisions/

Nickatak:
# docs: Add public ADR series in docs/decisions/

**Branch:** `docs/public-adrs` stacked on `refactor/error-envelope` (#745), which is the tip of the pattern PR chain (#739-#745), which is stacked on the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This PR adds a contributor-facing Architecture Decision Record series at `docs/decisions/`, capturing the reasoning behind the rewrite's non-obvious choices in one place. It is pure documentation - no code changes.

---

## Summary

13 ADRs covering the substantial decisions made across the modernization and pattern-PR chain. Reading order in the index goes foundation → process → patterns:

- **Tooling (consolidated)**: ADR-0001 covers all version pins and tool choices in one document (ruff, mypy, bandit, ESLint 9 pin, Vitest 3 pin, Knip, Stylelint, Prettier, Django 6, Python 3.13, Node 24, psycopg3, Postgres 18, Next.js 16, react-hooks rule downgrade, pre-commit framework rewrite).
- **Foundational architecture**: 0002 Stage 1 skips PeopleDepot, 0003 Separate frontend/backend containers, 0004 CSS Modules without preprocessor, 0005 Vitest replaces Jest.
- **Process**: 0006 Five-prefix git convention, 0007 Docstring conventions.
- **Code patterns**: 0008 FBVs over generics, 0009 Read/Write serializer split, 0010 Per-resource test files, 0011 URL routing, 0012 Permission class shape, 0013 Error envelope shape.

14 files added (13 ADRs + 1 index README) + 7 files modified - small stale-reference corrections sweeping the modernization stack (Django 6, DRF 3.17, PG 18, Node 24, Python 3.13, TS 6) through `README.md` and the `docs/developer/` tree (`design-system.md`, `backend.md`, `backend-docstring-style.md`, `deployment-infra.md`, `devops.md`, `quickstart-guide.md`). 983 insertions, 12 deletions. Three commits.

| # | Subject |
|---|---------|
| 1 | docs: Add public ADR series in docs/decisions/ |
| 2 | docs: Correct TypeScript version reference in design-system.md |
| 3 | docs: Correct stale references in README and docstring style guide |

## Why this shape

[...]

PR #747, opened 2026-05-03
chore: Adopt Make as canonical task runner with working host + stage workflows

Nickatak:
# chore: Adopt Make as canonical task runner with working host + stage workflows

**Branch:** `chore/makefile-rewrite` stacked on `docs/public-adrs` (#746), which is stacked on the pattern PR chain (#739-#745), which is stacked on the modernization chain (#734-#738). **Retarget chain when each parent lands.**

This PR makes the repo-root `Makefile` the canonical surface for common dev commands (per ADR-0014). The previous Makefile was broken in several ways - referenced removed services, ran tests against the wrong app name, named `db-shell` for what was actually Django's REPL. This rewrite establishes namespaced targets, fixes the recipes so they actually run end-to-end on a clean host, adds a host-process workflow plus a local stage smoke test, and switches `make lint` to use a venv-pinned `pre-commit` instead of whatever's on the host.

## Summary

9 files changed, ~520 lines net (the bulk is `poetry.lock` reflowing for the new pre-commit dev dep). Two commits.

| # | Subject |
|---|---------|
| 1 | chore: Adopt Make as canonical task runner with working host + stage workflows |
| 2 | docs: Add ADR-0014 and rewrite quickstart-guide for Make workflow |

## Surfaces (commit 1)

Targets are namespaced by the surface they touch:

- **`local-*`** - host-process workflow for rapid iteration. `local-install` cleans `node_modules` / `.venv` first, uses `npm ci` to keep the lockfile alpine-compatible, and bootstraps the venv with `python3.13` explicitly so poetry doesn't fall back to a missing bare `python` binary. `BACKEND_RUN` sources `dev/dev.env` and overrides `SQL_HOST=localhost` so host Django reaches the published pgdb port instead of the docker-DNS `pgdb` name.
- **`docker-*`** - dev-stack lifecycle (`up` / `down` / `logs` / `watch`).
- **`db-*`** - schema + data ops, with `read`-based `yes` confirmation prompts on destructive paths (`db-reset`, `db-reset-hard`).
- **`stage-*`** - production-shaped local simulation. `stage-smoke` is the primary tool: builds, ups, polls django (`/api/healthcheck/`) and next (`/`), tears down, exits 0/1. `stage-up` / `stage-down` exist for debugging when smoke fails. All three fail fast with an actionable error if `stage/stage.env` is missing. Real stage deploy still lives in HfLA's `incubator` repo; `stage-smoke` only validates that the images and compose work locally.
- **`lint`** / **`install-hooks`** - `pre-commit` is now a backend dev dep (`pyproject.toml`), so `make lint` invokes the venv's `pre-commit 4.6.0` instead of whatever system pre-commit happens to be installed. Resolves a stale-system-pre-commit issue surfaced during testing (Ubuntu's `/usr/bin/pre-commit 2.17.0` rejects the `pre-commit` stage name that the config uses).
- **`local-check-prereqs`** verifies node 24 / python3.13 / poetry / docker and exits non-zero on any miss. Surfaces missing prerequisites before deeper recipes fail with confusing errors.

`make help` is `.DEFAULT_GOAL`; lists everything with one-line summaries.

## Companion changes (commit 1)

- `dev/dev.env.example` swept of dead vars (`MODE`, `DEVTOOL`, `ENVIRON`, `DATABASE` - all Webpack/Vite-era or Django scaffolding leftovers with zero reads). Adds `BACKEND_INTERNAL_URL=http://django:8000` so `next.config.ts` can drive `/api/*` and `/admin/*` rewrites in dev. `SQL_PASSWORD` placeholder typo (`<same as POSTGRES_POSTGRES_PASSWORD>`) fixed.
- `.pre-commit-config.yaml`: `default_language_version.node` switched from `"24"` to `system`. nodeenv's `v24` prebuilt download URL doesn't resolve; we already require Node 24 on the host as a prereq, so reusing it is the right call.
- `backend/pyproject.toml` adds `pre-commit = "^4.4.0"` as a dev dep; `backend/poetry.lock` refreshed.
- `backend/.python-version` deleted (stale `3.12.4` pin from the pre-rewrite era; project now requires `^3.13`).
- `local-clean` no longer swallows permission errors - replaced the silent `find ... 2>/dev/null || true` chain with `rm -rf` so root-owned residue from prior sessions surfaces clearly.

## Docs (commit 2)

ADR-0014 records the decision: namespace conventions, confirmation-prompt pattern, `make lint` delegating to `pre-commit` (one source of truth across local and CI), variables at top, `help` as default goal. Standard ADR shape; index entry under "Process" in `docs/decisions/README.md`.

`quickstart-guide.md` rewritten around `make` targets rather than raw `docker compose ...` invocations. New sections: Database operations, Running tests, Stage smoke testing, Shell access, Local utilities. Lint section leads with `make lint` and keeps the individual-tool fallback for iterating on a single tool.

Also fixes a stale 404 envelope example in the quickstart: was the pre-error-envelope shape (with `status_code` / `message` keys); now matches the actual `civic_exception_handler` shape per ADR-0013.

## Test plan

- [x] `make help` renders the categorized command reference.
- [x] `make local-check-prereqs` reports all green when the host has node 24, python3.13, poetry, docker.
- [x] `make local-install` (frontend + backend) succeeds end-to-end. `npm ci` keeps `package-lock.json` byte-identical to committed.
- [x] `make docker-up` builds and starts all three containers cleanly.
- [x] `make db-up` / `db-migrate` / `db-makemigrations` / `db-grant-test-db-perms` all green.
- [x] `make local-makemigrations` / `local-test-backend` (10 tests OK) / `local-test-frontend` (11 pass / 3 skipped) / `local-test` (parent) all green.
- [x] `make lint` runs the venv's pre-commit 4.6.0; all 12 hooks pass.
- [x] `make stage-smoke` builds stage images, brings up the stack, both probes return green, auto-tears down, exits 0. ~2 minutes from cold cache.
- [x] `make db-reset` / `db-reset-hard` confirm prompts work; volume destruction verified.
- [x] `make local-clean` / `local-kill-ports` / `docker-down` clean up cleanly.

Skipped (long-running or interactive; verified by inspection): `local-run-frontend`, `local-run-backend`, `local-superuser`, `docker-logs`, `docker-watch`, `docker-shell-*`.

## Follow-ups (not blocking)

- `stage/stage.env.example` carries the same dead `ENVIRON` and `DATABASE` vars `dev.env.example` did; cleanup deferred (tracked locally).
- Anyone running `pre-commit install` against the host (system) pre-commit may want to upgrade out of Ubuntu's apt 2.17.0; `make install-hooks` now uses the venv binary so the system version is no longer load-bearing for the project's hooks.
- If a "fast subset" lint target later (e.g. ruff check + eslint, no type checking) becomes useful, it's an additive `make lint-fast` on top of this baseline.


RSkuma (approved):
Looks good to me

